PR #54, opened 2025-08-12
Add ops needed for SD 1.5 and add some debug helpers

owenhilyard:
Right now, tossing it back to the CPU is faster than going through dlpack, at least for the small tensors I tested with in comfyui. I can make that a toggle as well if you would like.


gabrieldemarmiesse (changes requested):
Thanks a lot for the PR! A bit more cleaning and we should be good to go!

Relative to passing on the CPU rather than using dlpack, I'd prefer avoiding it at the moment as it's the responsability of Modular to make it faster. Furthermore, outputs might be re-used on the gpu, for example when graph breaks are present

PR #169, opened 2025-09-28
`aten.ceil()` function

msaelices:
`aten.ceil()` implementation and tests

As there is not a MAX's `ops.ceil()`, we use `-ops.floor(-input)` instead for floats

Note: we could use `ops.floor(input) + 1` as an alternative.


gabrieldemarmiesse (changes requested):
Thanks a lot, just one thing to change

PR #170, opened 2025-09-30
Optimize `aten.ceil()` with custom Mojo kernel

msaelices:
Replace the two-step operation (`-floor(-x)`) with a single efficient Mojo kernel implementation for improved performance.

Changes:
- Add `math.mojo` kernel with `CeilKernel` implementation
- Update `aten_ceil()` to use `max_ops.custom()` with the new kernel

### DIFF
--- a/torch_max_backend/mojo_kernels/math.mojo
+++ b/torch_max_backend/mojo_kernels/math.mojo
@@ -1,0 +1,33 @@
+from compiler import register
+from gpu.host import DeviceContext
+from gpu.id import block_idx
+from layout import LayoutTensor
+from math import ceil
+from runtime.asyncrt import DeviceContextPtr
+from tensor_internal import InputTensor, OutputTensor, foreach
+
+from utils.index import IndexList
+
+
+@compiler.register("ceil")
+struct CeilKernel:
+    @staticmethod
+    fn execute[
+        dtype: DType,
+        rank: Int,
+        //,
+        target: StaticString,
+    ](
+        output: OutputTensor[dtype = dtype, rank = rank],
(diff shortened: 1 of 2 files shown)


torch_max_backend/mojo_kernels/math.mojo
```diff
@@ -0,0 +1,33 @@
...
+from layout import LayoutTensor
+from math import ceil
+from runtime.asyncrt import DeviceContextPtr
+from tensor_internal import InputTensor, OutputTensor, foreach
+
+from utils.index import IndexList
+
+
+@compiler.register("ceil")
+struct CeilKernel:
```
gabrieldemarmiesse, line 13: Thanks a lot for the PR! That's very cool to see custom op support for those new ops!

I talked offline with some modular devs about `foreach` and it seems there is something not very well documented:

>if I write an elementwise function with the foreach  function in mojo, will that be fused with other surrounding ops in the Max graph, like ops defined in the modular/ repo? 
> No, foreach does not fuse as far as I know. This is not a limitation of custom ops but of foreach. To make the kernel fuse, fn elementwise should be used to define custom elementwise ops instead of fn execute with foreach. Examples are in https://github.com/modular/modular/blob/main/max/kernels/src/Mogg/MOGGKernelAPI/MOGGKernelAPI.mojo#L626

So it seems we must avoid foreach and use `ElementwiseUnaryOp` to have this op being fused with the surrounding ops. Also if you have the time later on, i'd be cool to also change the other custom ops we have to use this pattern :)

gabrieldemarmiesse (changes requested):

gabrieldemarmiesse (approved):
Thanks a lot!

PR #182, opened 2025-10-01
`aten.cosh()` operation

msaelices:
- Implementation based on the `max_ops.exp()` op
- Tests

### DIFF
--- a/tests/test_aten_functions.py
+++ b/tests/test_aten_functions.py
@@ -585,6 +585,99 @@ def test_aten_ceil_scalar_tensor(device: str):
     check_functions_are_equivalent(fn, device, [x])
 
 
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64, torch.bfloat16])
+def test_aten_cosh_basic(device: str, dtype: torch.dtype):
+    """Test aten.cosh basic functionality with floating point numbers"""
+    # Skip float16 on CPU as MAX doesn't support f16 on CPU
+    if device == "cpu" and dtype == torch.float16:
+        pytest.skip("float16 not supported on CPU in MAX")
+
+    def fn(x):
+        return aten.cosh(x)
+
+    # Test with positive, negative, and zero values
+    # cosh(0) = 1, cosh is even function: cosh(-x) = cosh(x)
+    x = torch.tensor([-2.0, -1.0, -0.5, 0.0, 0.5, 1.0, 2.0], dtype=dtype, device=device)
+    check_functions_are_equivalent(fn, device, [x])
+
+
+def test_aten_cosh_special_values(device: str):
+    """Test aten.cosh with special mathematical values"""
+
+    def fn(x):
+        return aten.cosh(x)
+
+    # Test cosh(0) = 1 and symmetric property
+    x = torch.tensor([0.0, 1.0, -1.0, 2.0, -2.0], dtype=torch.float32, device=device)
+    check_functions_are_equivalent(fn, device, [x])
+
+
+def test_aten_cosh_2d_tensor(device: str):
+    """Test aten.cosh with 2D tensor"""
+
+    def fn(x):
+        return aten.cosh(x)
+
+    x = torch.tensor(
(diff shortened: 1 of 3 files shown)


msaelices:
@gabrieldemarmiesse, as the `sinh()` and `cosh()` functions are very similar, not sure if you prefer to reuse the values and tests for checking both `cosh()` and `sinh()` at the same time, something like this:

```python
@pytest.mark.parametrize("dtype", [torch.float32, torch.float64, torch.bfloat16])
def test_aten_sinh_and_cosh_basic(device: str, dtype: torch.dtype):
    """Test aten.cosh basic functionality with floating point numbers"""
    # Skip float16 on CPU as MAX doesn't support f16 on CPU
    if device == "cpu" and dtype == torch.float16:
        pytest.skip("float16 not supported on CPU in MAX")

    def cosh_fn(x):
        return aten.cosh(x)

    def sinh_fn(x):
        return aten.cosh(x)

    x = torch.tensor([-2.0, -1.0, -0.5, 0.0, 0.5, 1.0, 2.0], dtype=dtype, device=device)
    check_functions_are_equivalent(cosh_fn, device, [x])
    check_functions_are_equivalent(sinh_fn, device, [x])
```

tests/test_aten_functions.py
```diff
@@ -1039,6 +1039,170 @@ def fn(x):
...
+    x = torch.tensor(1.0, dtype=torch.float32, device=device)
+    check_functions_are_equivalent(fn, device, [x])
+
+
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64, torch.bfloat16])
+def test_aten_sinh_basic(device: str, dtype: torch.dtype):
+    """Test aten.sinh basic functionality with floating point numbers"""
+    # Skip float16 on CPU as MAX doesn't support f16 on CPU
+    if device == "cpu" and dtype == torch.float16:
+        pytest.skip("float16 not supported on CPU in MAX")
```
gabrieldemarmiesse, line 1129:
```suggestion
        pytest.xfail("float16 not supported on CPU in MAX")
```
Let's use an xfail instead since it should work. I serves as a reminder that we should remove the xfail at some point.

tests/test_aten_functions.py
```diff
@@ -1039,6 +1039,170 @@ def fn(x):
     check_functions_are_equivalent(fn, device, [x])
 
 
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float64, torch.bfloat16])
+def test_aten_cosh_basic(device: str, dtype: torch.dtype):
+    """Test aten.cosh basic functionality with floating point numbers"""
+    # Skip float16 on CPU as MAX doesn't support f16 on CPU
+    if device == "cpu" and dtype == torch.float16:
+        pytest.skip("float16 not supported on CPU in MAX")
```
gabrieldemarmiesse, line 1047:
```suggestion
        pytest.xfail("float16 not supported on CPU in MAX")
```

gabrieldemarmiesse (changes requested):
Thanks a lot for the PR!

gabrieldemarmiesse:
i don't think it's necessary. We get better failure reporting if those are separate tests. I'd be opened to parametrize if on the function with `pytest.mark.parametrize("func", [torch.cosh, torch.sinh])` if it's ok for you. We can even maybe throw tanh in this parametrize too

msaelices:
> i don't think it's necessary. We get better failure reporting if those are separate tests. I'd be opened to parametrize if on the function with `pytest.mark.parametrize("func", [torch.cosh, torch.sinh])` if it's ok for you. We can even maybe throw tanh in this parametrize too.
> 
> It's useful to differentiate quickly "all cosh tests are failing" from "all float16 tests are failing" when running the test suite. We don't get this info quickly if you put all the checks in the test function body.

Let me try the `parametrize` option, as I'm seeing how quickly is the test file growing while finding very similar patterns on the tests being added

gabrieldemarmiesse:
If you want to do a big refactor, maybe let's do that in a separate PR?

PR #213, opened 2025-10-21
Add `_log_softmax()` op

msaelices:
Formula: `log_softmax(x) = x - max(x) - log(sum(exp(x - max(x))))`

This is more numerically stable than computing `log(softmax(x))` directly, as it prevents overflow from exp(large_values) and underflow from log(small_values).

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,6 +155,10 @@ You can get various information with the following environement variables:
 * `TORCH_MAX_BACKEND_VERBOSE=1` to display the graph(s) made by pytorch and various other information.
 * `TORCH_MAX_BACKEND_BEARTYPE=0` to disable type checking. By default, everything in the package is type-checked at runtime. But it may lead to errors when actually the code is valid (and the type hint is wrong). You can try disabling the type-checking then to see if the bug goes away. Feel free to open a bug report in any case! Type errors should never happen and are a sign of an internal bug.
 
+## Running the unit tests
+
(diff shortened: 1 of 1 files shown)


tests/test_aten_functions.py
```diff
@@ -2331,3 +2331,47 @@ def fn(x):
...
         check_functions_are_equivalent(fn, device, [x])
+
+
+# aten._log_softmax(Tensor self, int dim, bool half_to_float) -> Tensor
+@pytest.mark.parametrize("dtype", [torch.float32])
+@pytest.mark.parametrize("dim", [-1])
+def test_log_softmax_basic(device: str, dtype: torch.dtype, dim: int):
+    """Test _log_softmax basic functionality.
+
+    Note: Currently only testing CPU and limited CUDA configs due to MAX GPU reduction
```
gabrieldemarmiesse: Since you're using the `device` fixture, this test runs on the gpu AND cpu, this comment is wrong

gabrieldemarmiesse (changes requested):

tests/test_aten_functions.py
```diff
@@ -2331,3 +2331,73 @@ def fn(x):
         match="The implementation of aten.max_pool2d_with_indices doesn't support returning indices yet.",
     ):
         check_functions_are_equivalent(fn, device, [x])
+
+
+# aten._log_softmax(Tensor self, int dim, bool half_to_float) -> Tensor
+@pytest.mark.parametrize("dtype", [torch.float32])
+@pytest.mark.parametrize("dim", [-1])
```
gabrieldemarmiesse: you don't need a parametrize if there is only a single value

PR #237, opened 2025-10-27
MNIST Demo and Tests

msaelices:
Fully trained MINST model using max_backend.


msaelices:
> You're right, let's remove the test, it's too slow

done

PR #238, opened 2025-10-27
Migrate tests to use the `conf` pattern

### LINKED ISSUE #235: Finish up the conversion of the unit tests to use the `conf` pattern
We recently moved many tests to use the `conf` pattern instead of the `device` pattern. Notably because the `conf` pattern tests the eager mode, while the `device` pattern doesn't. There are still around 200 tests using `device: str`. We should find out why they can't be converted out of the box and then implement whatever is necesssary to make them work in eager mode with `conf: Conf`

msaelices:
Partially fixes #235 

There are a few tests that are failing if refactored to the `conf` pattern:
```
    test_native_batch_norm_legit_no_training_basic
    test_native_batch_norm_legit_no_training_different_channels
    test_native_batch_norm_legit_no_training_none_weight_bias
    test_native_batch_norm_legit_no_training_different_eps
    test_native_batch_norm_legit_no_training_2d_input
```
The error we get: `RuntimeError: We don't support returning the saved mean in aten._native_batch_norm_legit_no_training yet to Tensor`

### DIFF
--- a/tests/test_aten_functions.py
+++ b/tests/test_aten_functions.py
@@ -2288,29 +2288,29 @@ def test_aten_max_no_dim(conf: Conf):
 
 @pytest.mark.parametrize("dim", [0, 1, 2])
 @pytest.mark.parametrize("keepdim", [True, False])
-def test_aten_max_with_dim(conf: Conf, dim: int, keepdim: bool):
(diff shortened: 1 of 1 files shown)


gabrieldemarmiesse (approved):
Nice, thanks!!!

PR #239, opened 2025-10-28
Fix mnist backward bug

gabrieldemarmiesse:
tensor dimensions must be returned as python integers, not scalar tensors

### DIFF
--- a/torch_max_backend/torch_compile_backend/compiler.py
+++ b/torch_max_backend/torch_compile_backend/compiler.py
@@ -336,7 +336,7 @@ class _GraphFactory:
         # None outputs can be required. So we remember here if
         # we want an output tensor (and we reccord the tensor position)
         # or if we want None.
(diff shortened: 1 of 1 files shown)


torch_max_backend/torch_compile_backend/compiler.py
```diff
@@ -396,11 +425,13 @@ def __call__(self, *args) -> list[torch.Tensor | None]:
 
         # Reconstruct the original output structure with None values
```
msaelices: I think this comment is now misleading as we do only not return None

torch_max_backend/torch_compile_backend/compiler.py
msaelices: I think the typings are wrong
```suggestion
        output_blueprint: list[tuple[OutputBlueprintKind, int | None]] = []
```

PR #358, opened 2026-08-13
Fix bfloat16/float16 floor_divide rounding across integer boundaries

gabrieldemarmiesse:


gabrieldemarmiesse:
@gabrieldemarmiesse-ai-agent i'm going to close this PR, can you open another PR with this branch from your own fork? Also fix the conflicts. After you open the PR, run the bench to see if this change in floor_divide trigerred any slowdown.

gabrieldemarmiesse-ai-agent:
Done — reopened from my fork as #369, with `main` merged in and the conflicts resolved. Same two commits (`016281c`, `3654a39`) plus the merge.

The conflict turned out to be semantic rather than textual, so it's worth flagging what I changed. This branch wrote the fp32-widening fix **twice** — in `_bin_bcast_body` and again in its SIMD twin `_bin_vec_op` — because those were separate bodies. #355 has since unified them into one `_bin_vec_op[dtype, out_dtype, op_code, is_cmp, width]`, so **both functions this branch patched no longer exist**.

I resolved it by taking the unified body and porting the widening into its single `BOP_FLOORDIV` branch, mirroring the treatment `BOP_POW` already gets there. Net effect: the fix now lives in one place instead of two, and it also covers the vectorized kernels, which didn't exist when this branch was written.

Your regression test `test_fast_floor_divide_narrow_float_boundary` auto-merged unchanged and passes — that's what pins the boundary case. Since the vectorized path is new to this fix I also checked parity against `torch.floor_divide` on each route the merge now takes: 1M contiguous (vectorized flat kernel), a 7-element tail (scalar remainder), and a transposed operand (general strided kernel). All match, bf16 and fp16.

Benchmarks are running now. I'm running the whole binary family rather than just `floor_divide`, because the edit is inside `_bin_vec_op` and every binary op funnels through it — the new branch is a `comptime if` so it should compile away for other op codes, but that's worth demonstrating rather than assuming. I'll post the numbers on #369.

PR #364, opened 2026-08-13
Fix OpInfo: RNG in-place ops, dtypefallback skips, masked_fill validation

gabrieldemarmiesse:

PR #367, opened 2026-08-13
Fix OpInfo: RNG in-place ops, dtype-fallback skips, masked_fill validation

gabrieldemarmiesse-ai-agent:
Reopened from the fork at your request — supersedes #364, same two commits (`91e896d`), no content changes.

Three independent bugs surfaced by `conformance/test_opinfo.py::test_errors_match` in CI (group 3/20, 75 failed):

- `aten::uniform_`, `aten::random_`, `aten::random_.to`, `aten::random_.from` were unregistered on the mojo device, so `torch.testing.make_tensor` could not build integer/float sample tensors on-device at all (142 failures). Registered with the same CPU-fallback-then-copy pattern as `normal_`.
- `test_errors_match` omitted the NotImplemented-sentinel check upstream's own `test_ops.py::test_errors` uses to catch dunder ops (`__rmod__` etc.) invoked directly on two Python scalars (3 failures).
- Some `error_inputs_func` corpora build a deliberately wrong-dtype tensor (complex/float64) purely to provoke the error under test; MAX doesn't support those dtypes, so construction raises before the op runs. Broadened not-implemented detection and wrapped the generator iteration so this skips like any other backend limitation (7 failures).
- `masked_fill` silently accepted a non-0-dimensional `value` tensor by squeezing it, where aten requires exactly 0 dimensions and raises `RuntimeError` — a genuine correctness gap (1 failure).

The second commit fixes a swallowed-skip bug found in review of the first: `self.skipTest(...)` was reachable from inside `with self.assertRaises(error_input.error_type)`, so for any OpInfo declaring a broad `error_type`, `unittest.SkipTest` was absorbed as "the expected exception" and a not-implemented case reported a false PASS.

Verified on an H100 via kyutai_scavenge: group 3's `not mojo_device` split runs 84 passed, 249 skipped, 0 failed (previously 75 failed).

Branch is 9 commits behind `main` but merges cleanly.

https://claude.ai/code/session_01UtQQafXhqRNJ8DoSPDx1zn


conformance/test_opinfo.py
```diff
@@ -57,7 +57,16 @@ def _not_implemented(exc: BaseException) -> bool:
     if isinstance(exc, NotImplementedError):
         return True
     text = str(exc).lower()
-    return "not implemented" in text or "no fast implementation" in text
+    return (
```
gabrieldemarmiesse: actually i don't think we should use this function _not_implemented to know when something is not implemented on our side. The problem is that, if a PR, by mistake, raise a NotImplementedError or similar in a kernel, then the test won't be reported as failed. I think we should have a function here that returns the list of tests that we know can't pass (this list will likely change depending on the type of gpu we run on since some gpus might have some op already implemented but not some other). Adding support for something means removing an element from this list. The kernel shouldn't have the ability to avoid a fail at will.

gabrieldemarmiesse (changes requested):

conformance/test_opinfo.py
gabrieldemarmiesse-ai-agent: You were right, and removing the sniff exposed something worse than the failure mode you described.

**Implemented in `e558ee9`.** `conformance/known_unsupported.py` declares cases node-by-node, consulted *before* the call. A declared case still runs and must still fail — xfail-strict — so when one starts passing the suite fails loudly naming the entry to delete. An undeclared `NotImplementedError` now fails. `_not_implemented` is gone. The kernel can no longer excuse itself.

**`(op, dtype)` turned out to be insufficient**, which is worth recording: 29 of the 76 error-corpus entries belong to operators whose `test_matches_cpu` node *passes*, so keying on the operator alone would have muted working cases. The key is `(test, operator token, dtype token)` — the tokens of the node id.

**Per-GPU, as you predicted it would need to be.** Base tables measured on `BASE_ACCELERATOR` (`sm_90a`, read from the same `architecture_name` the kernels gate on), plus per-architecture deltas where one operator line replaces its base line, an empty tuple meaning "supported here". No second copy of the table per GPU.

**Regeneration is scripted**, since this list needs updating whenever an op lands: `conformance/regenerate_known_unsupported.py --write [-k op]`, which runs the suite as its own pytest plugin and merges per node. One trap it has to handle: OpInfo re-raises every failure as a plain `Exception`, so the original type is gone by report time and the script must walk `__cause__`.

### Numbers

Full suite, 2786 nodes: **15F / 973P / 1798S → 15F / 973P / 237S / 1561 xfail**. Same 15 failures, the excused set identical, and **zero unexpected passes** — so the declaration reproduces exactly what the sniff was excusing, without the sniff.

Narrow selection (`-k "bmm or amax or dot or masked_fill or fft_fft2 or tensor_split or empty or dropout"`): 1F/15P/110S → 1F/15P/45S/65xfail.

### The finding: 62 false passes

`NotImplementedError` **subclasses `RuntimeError`**. So in `test_errors_match`, every `assertRaises(RuntimeError)` was *satisfied* by an operator that simply does not exist — 62 cases were taking credit for correct error handling by being absent. I verified the root cause directly: `issubclass(NotImplementedError, RuntimeError)` is `True`.

That is strictly worse than the case you raised: the sniff hid absences as *skips*, but this hid them as *passes*. The new code blocks that credit, otherwise switching to declarations would have converted 62 silent skips into 62 silent passes.

No new genuine bug surfaced beyond that. `FINDINGS.md` sections 2 and 3 no longer reproduce, so I updated it with the new totals.

`tests/ -n 15`: 3116 passed, 1 failed (`test_fast_mm_degenerate_dims[cpu-float16]`, passes in isolation — parallel flake, unrelated path).

### Sequencing

The RNG work is on a separate branch and nothing was pushed there, so no conflict at push time. The coupling is measured rather than guessed: stripping the RNG registrations flips exactly 2 nodes (`test_errors_match_uniform`, `test_errors_match_nn_functional_rrelu`), repaired by `regenerate_known_unsupported.py --write -k "test_errors_match and (uniform or rrelu)"`.

PR #360, opened 2026-08-13
Fix eager-mode tan precision on GPU float32 (mojo device)

gabrieldemarmiesse:

PR #368, opened 2026-08-13
Fix eager-mode tan precision on GPU float32 (mojo device)

gabrieldemarmiesse-ai-agent:
Reopened from the fork at your request — supersedes #360, same 8 commits (head `098de8a`), no content changes. The original had an empty body; commits as they stand:

- `afc0608` Fix eager-mode tan precision on GPU float32 (mojo device)
- `6abc028` Apply ruff format to fit_tan_poly.py
- `7bd9888` Address code review: honest tan accuracy claims, fix validation fidelity
- `02ec9a0` Merge branch 'main' into fix-tan-gpu-approx-precision
- `a0c28a5` Remove
- `28e9789` Merge branch 'main' into fix-tan-gpu-approx-precision
- `ff17748` Extract tan routing into op_utils.custom_tan
- `098de8a` Point data_movement_ops at the renamed _device_sm_count

Branch is 1 commit behind `main` and merges cleanly.

https://claude.ai/code/session_01UtQQafXhqRNJ8DoSPDx1zn


gabrieldemarmiesse:
can you update the branch?

gabrieldemarmiesse-ai-agent:
Updated — merged `main` in, head is now `eaa81e5`, 0 behind, still `MERGEABLE`. Used GitHub's own update-branch so it's a merge commit, consistent with the two already on this branch rather than a rebase that would rewrite your history.

CI re-triggered on the new head (run 31722467177) — all jobs pending as of now.

#367 is 9 behind `main` (also conflict-free) — want the same treatment there?

torch_mojo_backend/eager_kernels/op_utils/__init__.mojo
```diff
@@ -93,6 +95,104 @@ def ieee_sqrt[
...
+# `cos.approx.ftz.f32`), and dividing two such values right where cos(x) is
+# small (near a pole, x ~ pi/2 + k*pi) turns that fixed absolute error into a
+# large *relative* error in the quotient -- randn inputs land in that band
+# often enough to fail conformance (8.8% of a 20x20 sample).
+#
+# That argument is NVIDIA's; AMD and Apple reach `sin`/`cos` through
+# `llvm.sin` / `llvm.air.sin` instead, which are not the approx instruction.
+# The float32 GPU route is still the polynomial on every backend, both because
+# it is what those backends already ran and because it needs no per-vendor
+# claim about how accurate their `sin`/`cos` are.
```
gabrieldemarmiesse: This comment is way too verbose, make it smaller.

torch_mojo_backend/eager_kernels/data_movement_ops/data_movement_ops.mojo
```diff
@@ -2283,7 +2283,7 @@ def _repeat_tiled[
     var cout = ecols * r1
     var nout = erows * ncopies  # output rows, copies folded in
     var wide = ecols * size_of[dtype]() >= SEG_MIN_BYTES
-    var sm_count = _runtime_sm_count(ctx)
+    var sm_count = _device_sm_count(ctx)
```
gabrieldemarmiesse, line 2286: Is this a good idea? I remember a PR changed this to runtime sm count because it was more accurate for some gpus. Check in the git history you'll find something

gabrieldemarmiesse (changes requested):

torch_mojo_backend/eager_kernels/op_utils/__init__.mojo
gabrieldemarmiesse-ai-agent: Shortened to 9 lines in `4b474c8` — kept the routing, the approx-instruction cause, the pole, the 8.8%-of-a-20x20-sample measurement, and the one line on why AMD/Apple take the polynomial anyway. Dropped the `_call_libm` `comptime assert` detail and the `llvm.sin`/`llvm.air.sin` specifics, which are discoverable from the code.

torch_mojo_backend/eager_kernels/data_movement_ops/data_movement_ops.mojo
gabrieldemarmiesse-ai-agent: Good instinct, and worth checking — but the runtime query is preserved. `_device_sm_count` **is** the runtime query:

```mojo
comptime fallback = ctx.default_device_info.sm_count
var count = 0
try:
    count = ctx.get_attribute(DeviceAttribute.MULTIPROCESSOR_COUNT)
except:
    count = fallback
return count if count > 0 else fallback
```

I diffed the two as they existed side by side at `76f8bfd`, and the executable bodies are **identical** — only the docstring prose differs. Both query `MULTIPROCESSOR_COUNT` at runtime and use the comptime table purely as a fallback, so the H100 PCIe 114 / SXM 132 distinction you're remembering still works.

The history is that they were duplicates, not alternatives: `_runtime_sm_count` arrived with #349 (reductions), `_device_sm_count` with #350 (batchnorm), each solving the same problem in its own PR. #355 collapsed them and kept `_device_sm_count`.

**But the more important part: this line isn't optional, it's a fix for a broken `main`.**

#355 deleted `_runtime_sm_count`. #357 (repeat) added callers of it. Different files, so git merged both cleanly — and the result does not compile:

```
data_movement_ops.mojo:57:5: error: package 'op_utils' does not contain '_runtime_sm_count'
```

On current `main`, `aten::repeat` fails with that ImportError, and so does anything else in `data_movement_ops` — I hit it on `test_where` and `test_masked_fill` while benchmarking. It's nasty to notice because the build failure is **deferred**: my first probe called `x.repeat(2, 2)` and printed the right shape, because shape is metadata and nothing drained the queue. The error only surfaces on sync.

So `098de8a` is what makes `data_movement_ops` compile again. Without it the module is dead on `main`.

One fair criticism though: if the name made you read it as the compile-time table, the name is the problem. `_device_sm_count` reads like `default_device_info`, while `_runtime_sm_count` said exactly what it does. Happy to rename it back to `_runtime_sm_count` (or `_queried_sm_count`) across both call sites if you'd prefer — it's a pure rename, and the surviving name was an arbitrary choice when the duplicates were merged.

PR #369, opened 2026-08-14
Fix bfloat16/float16 floor_divide rounding across integer boundaries

gabrieldemarmiesse-ai-agent:
Reopened from my fork at @gabrieldemarmiesse's request, replacing #358, with `main` merged in and the conflicts resolved.

Same two commits as #358 (`016281c`, `3654a39`) plus the merge.

### What the fix does

`floor_divide` on bf16/fp16 did the division in the narrow dtype. With an 8- or 11-bit mantissa the quotient can round onto the wrong side of an integer boundary *before* the floor is taken — a true quotient of 5.985 becomes 6.0, so `floor_divide` returns 6 where torch returns 5. The fix widens the division to fp32 and narrows once at the end, which reproduces torch's output (torch's own `div_floor_floating` in `c10/util/generic_math.h` is more involved than divide-then-floor, but matches empirically on real bf16/fp16 tensors including that boundary case).

### The conflict was semantic, not textual

The branch had to write the fix **twice** — once in `_bin_bcast_body` (scalar) and once in its SIMD twin `_bin_vec_op` — because those were separate bodies. #355 has since unified them into a single `_bin_vec_op[dtype, out_dtype, op_code, is_cmp, width]` that every kernel in the family funnels through, so **both functions this branch patched no longer exist**.

Resolved by taking the unified body and porting the fp32 widening into its one `BOP_FLOORDIV` branch, mirroring the identical treatment already given to `BOP_POW`. Two consequences worth noting:

- the fix now exists **once** instead of twice, and
- it reaches the **vectorized** kernels, which did not exist when this branch was written.

Narrowing goes straight to `out_dtype` (equal to `dtype` for every non-comparison op) rather than to `dtype` and then storing, so there is no double narrowing.

### Verification

- `test_fast_floor_divide_narrow_float_boundary` — the regression test this branch added — auto-merged unchanged and passes. That is what pins the boundary case.
- Checked parity against `torch.floor_divide` on each path the merge now routes through, since the vectorized one is new to this fix: 1M contiguous (vectorized flat kernel), a 7-element tail (scalar remainder), and a transposed operand (general strided kernel). All match, bf16 and fp16.
- Benchmarks: running, results to follow in a comment.


gabrieldemarmiesse-ai-agent:
### Benchmarks: no slowdown from the floor_divide change

`benchmarks/test_binary.py`, whole binary family rather than just `floor_divide` — the edit is inside `_bin_vec_op` and every binary op funnels through it.

**124 passed, 0 regressions.** No node reported a `performance regression` or a `too noisy to trust` line. `floor_divide` itself is on baseline for both dtypes and both shapes.

That's the expected outcome and worth stating why: the added branch is a `comptime if dtype == float16 or dtype == bfloat16`, so for every other dtype it compiles away entirely, and for bf16/fp16 it replaces one narrow-dtype divide with an fp32 divide plus two casts — arithmetic that is free next to the memory traffic on a bandwidth-bound elementwise kernel.

### 8 failures, all pre-existing on `main` and unrelated to this PR

```
test_where[C_4096x4096-bf16]        test_masked_fill[C_4096x4096-Scalar-f32]
test_where[C_4096x4096-f32]         test_masked_fill[C_4096x4096-Tensor-f32]
test_where[A_357x789-bf16]          test_masked_fill[A_357x789-Scalar-f32]
test_where[A_357x789-f32]           test_masked_fill[A_357x789-Tensor-f32]
```

All eight are the same ImportError, not a timing result:

```
mojo build failed for data_movement_ops:
  data_movement_ops.mojo:57:5: error: package 'op_utils' does not contain '_runtime_sm_count'
```

`main` cannot build `data_movement_ops` at all. #355 deleted `_runtime_sm_count`, #357 added callers of it, different files so git merged both cleanly into something that does not compile. It reproduces on `main` with no changes of mine, and #368 already contains the fix (`098de8a`, pointing those call sites at `_device_sm_count`, which does the same runtime `MULTIPROCESSOR_COUNT` query). Details on that thread: #368 (discussion_r3782674887).

Worth knowing that the failure is **deferred** — my first probe called `x.repeat(2, 2)`, got the correct shape back and looked fine, because shape is metadata and nothing drained the kernel queue. It only surfaces on sync, which is why it slipped through two merges.

Nothing here blocks this PR; flagging it because it makes `main` red for anything touching `data_movement_ops` until #368 lands.

PR #372, opened 2026-08-14
Do not block the first device tensor on one-time runtime setup

### LINKED ISSUE #371: Do not block the code when compiling a tensor creation on the gpu
Here is a code snippet, to try with a cold cache (no .so, no mojo cache):

```python
import torch
import torch_mojo_backend

torch_mojo_backend.register_mojo_devices()

print("------>  defining tensors")

a = torch.tensor([1, 2, 3, 4], device="mojo:0")
b = torch.tensor([10, 2, 3, 4], device="mojo:0")
c = torch.tensor([100, 2, 10, 4], device="mojo:0")

print("------>  tensors defined")

d = (a + b - c) * 8 / 16

print("------>  calculation declared")

print(d)
```

You'll notice it blocks when creating `a` instead of progressing. Then when it's done, it compiles all the ops needed for `(a + b - c) * 8 / 16` at the same time (none of those ops being compiled blocks the progression). Declaring a tensor on the gpu should not make the compilation blocking. It should behave the same as all the other ops and return directly while enqueuing the op to be compiled in the background.

gabrieldemarmiesse-ai-agent:
Closes #371.

### It is not compilation

That's what the asymmetry suggests, so I profiled it. With the extension already loaded, **0.711 of the first tensor's 0.721 s sits inside a single native call** — `tensor_holder.alloc_from_host` — and the second tensor takes 0.000 s:

```
mojo_device__to_copy                    0.721
  _from_cpu                             0.711
    tensor_holder.alloc_from_host       0.711   <- tottime, one call
```

That is one-time **runtime** setup (device context + the pinned-host staging path) landing on whoever allocates first. Operations look non-blocking only because they are never first — they'd pay it too.

### Two one-time costs, both now started at registration on a background thread

1. building/dlopening the `tensor_holder` extension (~2.4 s cold), which `eager_kernels.tensor_holder`'s module `__getattr__` otherwise does **synchronously on the calling thread**;
2. the `alloc_from_host` setup above, warmed with a 1-element tensor.

The second must go through `_from_cpu`: prewarming with a plain `alloc` measurably did **not** help — the two entry points don't share the setup.

### Measured

| | before | after |
|---|---|---|
| first tensor, warm cache, 1.5 s of startup work first | 0.718 s | **0.007 s** |
| first tensor, warm, nothing before it | 0.72 s | 0.72 s |

The second row is the honest limit: with nothing between `register_mojo_devices()` and the first tensor — the shape of the snippet in the issue — the wall clock is unchanged, because the work still has to happen and there is nothing to overlap it with. Any real program (build a model, load data, read a checkpoint) is the first row.

### Why not return immediately with shape/dtype/strides instead

Because for an operation the queue defers the *launch* but **preallocates the output** — `call_queue.kernel_call_into` is documented as "a descriptor call whose outputs were preallocated in Python", and its `args` are *"raw pointers and scalars"*, with `keepalive` holding the tensors only so the buffers survive. The pointer is baked into the queued launch args **at enqueue time**.

So a tensor cannot hand out a pointer later without the queue holding the *tensor* instead of the pointer and re-reading `_ptr` at drain. That is a real change, and it has two further consequences worth weighing before anyone attempts it:

- OOM would surface at drain rather than at the allocation site, losing the traceback that points at the offending line;
- `_alloc_with_recovery` currently recovers from OOM by **draining the queue** to reclaim memory and retrying once. If allocations themselves happen during the drain, that recovery is circular.

Given the prewarm already covers every program that does anything before its first tensor, I did not think that trade was worth making unprompted — happy to if you want it.

### Details

`register_mojo_devices()` is the hook rather than import: it is where a program declares it will use the device, and importing the package must keep compiling nothing, which torch.compile-only users depend on. Skipped entirely under `TORCH_MOJO_BACKEND_KERNEL_QUEUE=0` so that kill switch keeps meaning "every build blocks at its call site". Daemon thread; the exception is swallowed because `_ensure_tensor_holder` caches a failure and re-raises it at the real call site, so a broken build still surfaces there with an actionable traceback.

Two tests, asserting observable state rather than timings so they cannot flake on a loaded machine: the holder ends up loaded without the test creating a tensor, and no thread starts when the queue is disabled. `tests/test_mojo_device_runtime.py` + `test_mojo_device.py` + `test_call_queue.py`: 159 passed, 6 skipped, 4 xfailed.

### DIFF
--- a/tests/test_mojo_device_runtime.py
+++ b/tests/test_mojo_device_runtime.py
@@ -1,13 +1,14 @@
 """PrivateUse1 device-module, transfer, and RNG runtime contracts."""
 
 import gc
+import time
 import weakref
 from concurrent.futures import ThreadPoolExecutor
 
 import pytest
 import torch
 
-from torch_mojo_backend import register_mojo_devices
+from torch_mojo_backend import get_accelerators, register_mojo_devices
 from torch_mojo_backend.mojo_device import torch_mojo_tensor as mojo_tensor_module
 from torch_mojo_backend.mojo_device.torch_mojo_device_module import (
     _reserve_philox_state,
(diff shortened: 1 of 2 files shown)


gabrieldemarmiesse-ai-agent:
Closing as asked — you're right that this was the wrong fix. `tensor_holder.full` reads like it's compiling an `aten.full` kernel, but `.full` is just the specialization tag for an ungated build, and the thing being compiled is the metadata holder. Nothing anomalous was happening; the message made it look like there was.

Doing the real fix on a fresh branch: name the extension properly in the trace, drop the `t=` stamps, and report elapsed compile time on the `built` line instead.

PR #362, opened 2026-08-13
Skip cross-device comparison where OpInfo says output may differ

gabrieldemarmiesse:

PR #374, opened 2026-08-14
Skip cross-device comparison where OpInfo says output may differ

gabrieldemarmiesse-ai-agent:
Replaces #362, reopened from the fork at the owner's request (original PR was on a branch on this repo directly; #362 was closed without merging). Same 5 commits, merged up to date with `main` (now at #373).

## What the 5 commits do

`conformance/test_opinfo.py::test_matches_cpu` compares CPU vs mojo output values for every OpInfo sample. For a set of operators that comparison is fundamentally invalid, and PyTorch's own OpInfo metadata already says so via a `TestCommon.test_compare_cpu` skip on the op's `OpInfo` entry (uninitialized memory for `empty*`, a separate RNG stream for `dropout`/`multi_head_attention_forward`, and `native_batch_norm`'s well-documented CPU/CUDA `save_mean`/`save_invstd` shape split, pytorch/pytorch#85960). Rather than hand-maintain a duplicate "known-divergent" list that would drift from upstream's own, the branch adds `_cross_device_comparison_skip_reason()`, which inspects `op.skips` for that same `DecorateInfo` and honors it the same way upstream's own consistency test would.

1. **Skip cross-device value comparison where OpInfo itself says output may differ** — the core change described above, plus a separate fix: `tensor_split`'s indices/sections argument must stay on CPU by API contract on every backend, so the harness's blind "move every tensor to device" strategy is recognized as a harness limitation (`_harness_cannot_construct`) and skipped rather than treated as a real failure.
2. **Catch not-implemented errors raised while moving samples to device, too** — the try/except previously only wrapped the op call, not the `sample.transform(...)` device-move step itself; `sparse_sampled_addmm`'s sample triggers an unimplemented op during the move, before the op under test is ever reached.
3. **Document the native_batch_norm test_compare_cpu case in detail** — records the concrete ATen source references (CPU vs CUDA kernels) so a future reader doesn't have to rediscover them.
4. **Guard `_cross_device_comparison_skip_reason` against non-DecorateInfo skips** — `op.skips` isn't homogeneous; `linalg.solve_triangular` and `lu_unpack` mix in a bare decorator function (`skipCPUIfNoLapack`) alongside proper `DecorateInfo` instances, which previously raised `AttributeError` instead of "no skip found."
5. **Harden skip-reason logic found by code review** — declines a callable `active_if` instead of treating it as unconditionally active by truthiness (no current skip uses this, but it was a latent gap), and stops conflating the `tensor_split` harness-limitation skip with a real `NotImplementedError` message.

## Before / after: conformance impact

A full `conformance/` sweep is ~40 minutes and 2786 nodes; since this change only *adds skips*, the informative measurement is the skip/fail delta on exactly the operators the change touches, run identically on both a `main` baseline worktree and this branch:

- Selection: all 61 operators for which `_cross_device_comparison_skip_reason` fires (computed directly from `op_db`, not guessed), plus `tensor_split` and `sparse_sampled_addmm` (handled by the separate try/except-ordering fix) — 188 `test_matches_cpu` node IDs in total, via `-k`.

| | before (`main` @ 6fccc00) | after (this branch) |
|---|---|---|
| failed | 47 | **0** |
| passed | 26 | 0 |
| skipped | 115 | 188 |

All 47 previously-failing nodes in this selection now skip (the intended effect). The 26 previously-*passing* nodes also move to skip: the OpInfo `test_compare_cpu` skip this branch honors is unconditional per (op, dtype) — same as upstream's own — not conditioned on whether a given sample happened to coincidentally match, so a handful of samples that passed by chance are correctly no longer asserted on either. 115 + 47 + 26 = 188, and 115 + 73 = 188 checks out against the after total. No errors in either run.

The remaining ~220 conformance failures outside this selection are pre-existing findings documented in `conformance/FINDINGS.md` (unimplemented ops, FFT/complex-dtype value mismatches, etc.) — unrelated to this change and not regressions.

## tests/

`uv run pytest tests/ -n 15` on this branch: **3101 passed, 1 failed, 41 skipped, 95 xfailed, 1 xpassed** (876s). The one failure, `test_aten_squeeze_5d[cpu]` (a MAX graph-compile failure with a generic "file an issue" message, in a test this branch never touches), reproduced as **2 passed, 0 failed** when rerun in isolation outside the `-n 15` parallel run — consistent with a compile-cache contention flake under heavy parallel MAX compilation, not a regression from this branch.

## Why #362 was UNSTABLE

Not this PR's doing: CI's `test` matrix job runs a bare `uv run pytest --splits N -k "..."` from the repo root, which (per `AGENTS.md`'s own warning) also collects `conformance/`, so the ~220 pre-existing conformance findings surface as CI failures on essentially every PR right now. The two most recent CI runs on `main` itself (`1aa3ea8`, `4cfefd8`) show the identical failing-shard pattern, confirming this is a pre-existing, repo-wide CI state rather than something introduced here.


conformance/test_opinfo.py
```diff
@@ -71,16 +142,21 @@ def test_matches_cpu(self, device: str, dtype: torch.dtype, op: Any) -> None:
...
-            )
             try:
+                moved = sample.transform(
+                    lambda x: x.to(device) if isinstance(x, torch.Tensor) else x
+                )
                 actual = op(moved.input, *moved.args, **moved.kwargs)
             except Exception as exc:  # noqa: BLE001 - triaging is the point
                 if _not_implemented(exc):
                     self.skipTest(f"not implemented on mojo: {exc}")
+                if _harness_cannot_construct(exc):
```
gabrieldemarmiesse: can we catch this earlier rather than just parsing the exception? Any change we could detect this before the exception?

gabrieldemarmiesse (changes requested):

conformance/test_opinfo.py
gabrieldemarmiesse-ai-agent: Replaced the message parse with a placement probe in `4fb2e67`, and it turned out better than parity.

Rather than tightening the regex, the harness now asks OpInfo where each sample tensor *belongs*: it builds the same samples a second time with `device="meta"` and reads back where each one landed. `sample_inputs_tensor_split` hardcodes `torch.tensor([1,2,3])` for the indices while the operand goes through `make_tensor(device=device)`, so the pinning is directly observable rather than inferred. `_harness_cannot_construct` is deleted.

`meta` is the right probe target specifically because it allocates nothing and runs no kernel — so it cannot fail for the reason samples are CPU-built in the first place, which is what ruled out probing the mojo device. Placement is trusted only when both builds line up tensor-for-tensor on shape and dtype; otherwise everything moves, exactly as before (~100 generators raise on meta, and `sort`/`argsort`/`to` diverge, so the fallback matters).

**The payoff is that a skip becomes a real comparison.** With the indices correctly left on CPU the call is legal, so `tensor_split` is now actually checked against CPU instead of skipped:

- 32-node selection: before 2F/10P/20S -> after 2F/**15P**/15S
- 296-node selection (chosen to exercise every fallback path): 15F/56P/225S -> 15F/61P/220S, and the node-by-node diff is *only* those 5 `tensor_split` SKIPPED->PASSED

I verified that independently: `pytest conformance/ -k tensor_split` gives **5 passed** on this branch. The probe is value-neutral (sample bytes identical across all 2573 op/dtype pairs) and costs ~4 ms/node.

Two judgement calls worth your eye:

1. The message parse is gone outright, so a future op that both fails `meta` construction *and* pins an argument will now fail loudly rather than skip. I think that is the right default — a loud failure is a finding — but it is a choice.
2. `__getitem__`/`index_put` indices now stay on CPU (ATen's `make_info` moves them itself). Invisible today since both are unimplemented skips, but our kernels will first meet that shape when they land.

Also noticed: `--reference-inputs` is declared in `conformance/conftest.py` but never consumed. If it gets wired up, `logsumexp` pins a dim tensor there too.

PR #359, opened 2026-08-13
Fix bfloat16 remainder producing wrong values near boundary divisors

gabrieldemarmiesse:

PR #375, opened 2026-08-14
Fix bfloat16/float16 remainder producing inf/wrong values near boundary divisors

gabrieldemarmiesse-ai-agent:
Reopened from my fork at @gabrieldemarmiesse's request, replacing #359, with `main` merged in and the conflicts resolved.

Same two commits as #359 (`0378260`, `4717e04`) plus the merge. Net diff vs `main` is two files: `tests/test_eager_kernels.py` (+29) and `torch_mojo_backend/eager_kernels/logic_ops/logic_ops.mojo` (+180/-2).

## What the fix does

`RemainderSpec`'s `%` ran the textbook `x - trunc(x / y) * y` decomposition at the operand's own precision. Forming the quotient explicitly has two separate failure modes:

- **bf16 overflows.** bf16 shares fp32's 8-bit exponent field, so a divisor near that format's smallest normal value (~1.18e-38 — e.g. what OpInfo's `exclude_zero` substitutes for an exact-zero draw) sends `x / y` to ±inf, even though the true remainder is always finite and bounded by `|y|`.
- **fp16 loses the low bits.** Independently of overflow, multiplying the quotient back by `y` in fp32 to recover `x` loses exactly the bits the remainder lives in once `|x / y|` exceeds fp32's 24-bit mantissa. fp16's own dynamic range reaches that with no boundary divisor in sight (23456.0 / 0.0009937… ≈ 2.36e7 > 2²⁴), where the old code silently rounded a true remainder of 0.0004178 down to exactly 0.

Both now route through a bit-exact fp32 `fmod` (ported from fdlibm/musl's `e_fmodf`) that never forms `x / y`: it aligns `y`'s mantissa to `x`'s exponent one bit at a time and subtracts, exact at every step. `__rmod__` shares the same kernel with swapped operands, so it is fixed by the same change.

## The conflict was semantic, not textual

The branch wrote the fix **twice** — once in `_bin_bcast_body` (the scalar body behind the general strided kernel and the CPU closure) and once in its SIMD twin `_bin_vec_op` (behind the vectorized kernels) — because those were separate bodies. #355 has since unified them into a single `_bin_vec_op[dtype, out_dtype, op_code, is_cmp, width]` that every kernel in the family funnels through, so **both functions this branch patched no longer exist**.

Resolved by taking main's unified body and porting the fix into its one `BOP_REMAINDER` branch, mirroring the identical treatment main already gives `BOP_POW` and `BOP_FLOORDIV`. Details worth a reviewer's eye:

- The fix now exists **once** instead of twice. Unlike #369, the set of routes it covers is unchanged: the vectorized kernels predate this branch and its second copy already served them. The scalar copy is now just `_bin_vec_op[..., 1]`.
- Narrowing goes straight to `out_dtype` (equal to `dtype` for every non-comparison op) rather than to `dtype` and then storing, so there is no double narrowing.
- The two new helpers (`_fmod_narrow_float_exact_scalar`, `_remainder_narrow_float_exact[width]`) merged unchanged. The only edit to them is the docstring reference to the now-deleted `_bin_bcast_body`, retargeted at `_bin_vec_op` instantiated at width 1.
- Both regression tests auto-merged, landing next to main's `test_fast_floor_divide_narrow_float_boundary`. The test-file conflict was purely textual (same insertion point).

## Verification

**The two regression tests this branch added pass**, on both the cpu and gpu mojo devices, alongside main's floor_divide regression test: `test_fast_remainder_bfloat16_near_zero_divisor`, `test_fast_remainder_float16_large_ratio` — 11 passed, 1 skipped for that selection. That is what pins the port.

**Per-route parity**, since one body now feeds every kernel in the family. bf16 and fp16, on: a 1M contiguous same-shape case (vectorized flat kernel, width 8); the same with tiny near-smallest-normal divisors sprinkled onto the vector path; lengths not a multiple of the vector width (7, 8193, 1000003) for the scalar tail; transposed lhs and transposed rhs plus a broadcast-expanded pair (general strided kernel); the `Scalar` / `Scalar_Tensor` overloads; and a sign/zero/inf quadrant sweep including division by zero. **Zero mismatches against an fp64 reference, on every route, both dtypes**, and our output is never NaN.

That fp64 reference is deliberate, and the reason is itself a data point for this PR: **`torch.remainder` on CPU bf16 is not a usable reference in exactly this regime.** Its vectorized CPU path hits the same overflow this PR fixes and returns NaN where the true remainder is finite — `torch.remainder(bf16(16.875), bf16(4.70e-38))` is `nan` in a 1M-element tensor but `0.0` for the same values as a 1-element tensor, and fp32/fp64/`math.fmod` all agree on `0.0`. In the two cases where I seeded tiny divisors densely, every single ours-vs-torch-CPU difference (214935 of 1048576, and 60 of 35904) is a position where torch-CPU produced NaN and we produced the correct finite value; against fp64 there are none. Worth knowing before anyone writes a bf16 remainder test against a large CPU tensor.

**`uv run pytest tests/ -n 15`: 3104 passed, 2 failed, 41 skipped, 95 xfailed, 1 xpassed (903s).** The two failures are pre-existing flakes of the parallel run, not regressions:

- `test_high_level_ops.py::test_layer_norm_basic[cpu]` — a MAX graph-compile failure with the generic "Failed to compile the model. Please file an issue" message, on the `torch.compile` path this PR does not touch.
- `test_mojo_device.py::test_non_blocking_h2d_does_not_drain_prior_gpu_work[gpu]` — a device-stream timing assertion.

Both **pass in isolation on this branch, and pass in isolation on a `main` worktree at `6fccc00`** (3 passed, 1 skipped). The layer_norm signature matches an independently-observed `-n 15` MAX compile-cache contention flake. Neither test's code path reaches `logic_ops`.

**Cross-compiled for `sm_90a` against `main`** with `scripts/compare_kernel_asm.py --modules logic_ops --ops RemainderSpec FloorDivSpec AddSpec --dtypes bfloat16 float16 float32 int64`: **exactly 6 of 36 kernels differ** — the three `RemainderSpec` kernels (`binary_flat_vec`, `binary_rowvec`, `binary_strided`) for bfloat16 and the same three for float16. `AddSpec` at every dtype, `FloorDivSpec` at every dtype, and `RemainderSpec` at float32 and int64 are byte-identical. That is exactly the intended blast radius.

No benchmark baselines touched.

## One thing a reviewer should weigh

`_remainder_narrow_float_exact[width]` calls the scalar fmod once per lane in a `comptime for`, and that scalar fmod is deliberately not `@always_inline` (its bit-serial loop is large). At the vectorized width that is 8 serialized calls per SIMD group with a data-dependent trip count, so bf16/fp16 `remainder` is slower than it was — correct instead of fast. The trade-off is unchanged from #359; I flag it because the unified body now makes it visible in one place, and there is no `benchmarks/` node for `remainder` to quantify it.


torch_mojo_backend/eager_kernels/logic_ops/logic_ops.mojo
```diff
@@ -298,8 +456,26 @@ def _bin_vec_op[
...
+            #
+            # One body serves every kernel in this family now, so this lives
+            # in one place; before the vectorization refactor the same fix had
+            # to be written twice, in the scalar body and its SIMD twin.
+            comptime if dtype == DType.bfloat16 or dtype == DType.float16:
+                return _remainder_narrow_float_exact[width](
+                    a.cast[DType.float32](), b.cast[DType.float32]()
+                ).cast[out_dtype]()
+            else:
+                return (a % b).cast[out_dtype]()
```
gabrieldemarmiesse: I think this if-else should be extracted in a function called `custom_remainder()`. You can move the big comment there

torch_mojo_backend/eager_kernels/logic_ops/logic_ops.mojo
```diff
@@ -298,8 +456,26 @@ def _bin_vec_op[
...
+            # the remainder's low bits outright, or (for bf16 specifically,
+            # whose exponent field is as wide as fp32's) overflowing to
+            # +/-inf when the divisor is near bf16's smallest normal value.
+            # Route both through the bit-exact fp32 fmod above.
+            #
+            # One body serves every kernel in this family now, so this lives
+            # in one place; before the vectorization refactor the same fix had
+            # to be written twice, in the scalar body and its SIMD twin.
+            comptime if dtype == DType.bfloat16 or dtype == DType.float16:
+                return _remainder_narrow_float_exact[width](
```
gabrieldemarmiesse: Do we need to rerun and update the benchmark now that we use this function instead of a%b?

gabrieldemarmiesse (changes requested):

torch_mojo_backend/eager_kernels/logic_ops/logic_ops.mojo
gabrieldemarmiesse-ai-agent: Done in `1182efe`. `custom_remainder` now sits in `op_utils/__init__.mojo:318`, immediately after `custom_tan` so the two dtype-routing numeric helpers live together; the big comment became its docstring and `BOP_REMAINDER` is a single call. The fmod helper moved with it and the SIMD wrapper folded in, so it is 3 functions down to 2.

Verified it is a pure refactor rather than trusting that it is: `compare_kernel_asm.py --modules logic_ops --dtypes float32 bfloat16 float16 int64` reports **0 of 488 sm_90a sidecars differ** across all 108 specializations. The extraction changes no emitted code.

torch_mojo_backend/eager_kernels/logic_ops/logic_ops.mojo
gabrieldemarmiesse-ai-agent: Yes — and the answer is not the one either of us wanted, so I want your decision rather than my judgement here.

Ratios ours/stock, H100 PCIe @1395MHz:

| node | main | this branch |
|---|---|---|
| f32 C_4096x4096 Scalar / Tensor | 1.022 / 0.998 | 1.021 / 0.992 |
| f32 A_357x789 Scalar / Tensor | 1.238 / 0.762 | 1.246 / 0.761 |
| **bf16** C_4096x4096 Scalar / Tensor | 1.003 / 0.990 | **3.16 / 3.45** |
| **f16** C_4096x4096 Scalar / Tensor | 1.005 / 0.992 | **3.24 / 3.53** |
| **bf16** A_357x789 Scalar / Tensor | 1.306 / 0.813 | **2.95 / 3.30** |
| **f16** A_357x789 Scalar / Tensor | 1.313 / 0.820 | **3.06 / 3.41** |

**f32 needs no baseline update** — max delta 0.6%, and the codegen is byte-identical, so it cannot have moved.

**bf16/f16 cost 2.3–4.2x.** Worth being precise about what that compares against: main's `1.003` is a *fast wrong answer* — bf16 remainder on main returns the wrong value near boundary divisors, which is the bug this PR exists to fix. So this is not a regression against a working implementation, it is the price of correctness. There was no bf16/f16 node in `benchmarks/` before (the changed regime was unbenchmarked, which is why nothing caught the bug or its cost), so I added the 8 nodes and recorded them; a read-only rerun reproduces them within 1.6%.

Whether 3x is an acceptable price for a correct `remainder` is your call, not mine. If it is not, the alternative is a cheaper correct algorithm — the current one routes through a bit-exact fp32 fmod, and there is likely something faster that is still exact for the boundary cases. Say the word and I will look rather than land this trade.

Two things found along the way that are **not** this PR's to fix, flagged so they are on record:

1. **f32 and f64 have the identical bug, unfixed.** `16.875 % 4.70e-38` returns `-inf` on mojo against `0.0` on CUDA. This PR only widens the 16-bit dtypes.
2. **The four f32 Scalar remainder baselines are stale on `main`.** They read 2.902/3.633 but measure ~1.24/1.02. Untouched here; worth a `--update-baselines=force` pass along with the GEMM ones.

gabrieldemarmiesse-ai-agent:
### Correction to this PR's description, and to something I said earlier

I previously doubted the claim that `torch.remainder` CPU bf16 is itself broken — I swept 40 trials x 4096 exact multiples and got zero NaNs, and said the claim should be discounted. **That was a badly designed probe, and the claim is real.** Reproduced exactly:

```python
import torch
def rem(n):
    x = torch.full((n,), 16.875,   dtype=torch.bfloat16)
    y = torch.full((n,), 4.70e-38, dtype=torch.bfloat16)
    return torch.remainder(x, y)

rem(16)[0]   # 0.0    <- correct
rem(31)[0]   # 0.0    <- correct
rem(32)[0]   # nan    <- vector path engages
rem(64)[0]   # nan
```

It needs **two** conditions simultaneously, which is why a naive sweep misses it: `n >= 32` (the bf16 vector width — below that torch stays on its scalar path and is correct) **and** a divisor small enough (~1e-38) that `x/y` overflows fp32. At 1e-37 there are zero NaNs. fp64 truth for those values is `1.2000087e-38`.

One correction to the claim as originally worded: the same thing happens for **f32** at those values, so this PR's description saying fp32 agrees on `0.0` holds only for a 1-element tensor.

### Three stale lines in the description

The `custom_remainder` extraction (`1182efe`) invalidated three statements above:

- there **is** now a `benchmarks/` node for remainder — 8 of them, bf16 and f16, added because that regime was previously unbenchmarked;
- `_remainder_narrow_float_exact[width]` no longer exists under that name; it moved into `custom_remainder` in `op_utils`;
- "narrowing goes straight to `out_dtype`" is still true but now happens inside `custom_remainder`, not in the `BOP_REMAINDER` branch.

Please read the two review-thread replies for the current state — in particular the benchmark one, which contains a correctness-versus-speed trade that needs your decision before this lands.

PR #379, opened 2026-08-14
One batched kernel body for the whole foreach elementwise family

### LINKED ISSUE #376: Improve the perf of the aten _foreach_mul_.Scalar
@gabrieldemarmiesse-ai-agent do it

gabrieldemarmiesse-ai-agent:
`aten::_foreach_mul_.Scalar` (#376) had **no batched path at all off Metal**:
`_foreach_metal_f32` declined unless `device.api == "metal"`, so it — plus
`_foreach_div_.ScalarList`, `_foreach_lerp_.Scalar`, `_foreach_addcmul_.Scalar`,
`_foreach_addcdiv_.ScalarList` and `_foreach_sqrt` — redispatched to ATen's
per-tensor decomposition on CUDA and ROCm.

Measured fresh before touching anything: **1.76x** stock on `L_4x1048576` and
**4.31x** on `L_16x65536` (H100 PCIe, device time). The gap is real, but the
recorded baselines (8.03 / 6.91) could not be reproduced at the commit that held
them — this PR re-records them.

## What changed

One kernel body, `foreach_batched_kernels.mojo`, comptime-parametrized by dtype
and by op (mul / add / div by a per-tensor host scalar, multiply by a 0-d device
scalar, lerp, addcmul, addcdiv, sqrt), reached through one Mojo bridge
`_foreach_ew_go[op]` and one Python launch helper. Six bridges collapse into it —
`ForeachScalarOp`, `ForeachLerpScalar`, `ForeachAddcOp`, `ForeachSqrt`,
`ForeachMulTensor` and `elementwise_ops`' `ForeachAddScalar` — along with the
hand-tuned `foreach_mul_tensor_f32_aligned_streaming_ilp8_t128_v8`. Net −406
lines. Apple regroups the *same* descriptors into the fixed-arity Metal launches,
which is the only part that ever needed to be Metal-specific.

**The Metal gate was not the whole cost.** The chunk was fixed at 65,536
elements, which put the 4x1M list on 64 blocks and the 16x64k list on 16 — of 114
SMs. It is now a runtime launch argument derived from the element count and the
runtime SM count (4 blocks/SM, swept: see the table next to the constant), so no
shape is baked in and every shape reuses one compiled kernel. That is where most
of the win is: `_foreach_add_.Scalar` and `_foreach_mul_.Tensor` already had a
single batched launch and still improved 2.2x and 2.4x.

## Measured (H100 PCIe, `benchmarks/test_foreach.py`, ours/stock device time, f32)

| op | 4x1048576 | 16x65536 |
|---|---|---|
| `_foreach_mul_.Scalar` | 1.76 → **0.833** | 4.31 → **0.387** |
| `_foreach_add_.Scalar` | 1.85 → **0.843** | 1.95 → **0.393** |
| `_foreach_div_.ScalarList` | 1.19 → **0.585** | 2.88 → **0.291** |
| `_foreach_lerp_.Scalar` | 4.55 → **0.682** | 6.88 → **0.439** |
| `_foreach_addcmul_.Scalar` | 2.43 → **0.832** | 4.55 → **0.434** |
| `_foreach_addcdiv_.ScalarList` | 2.36 → **0.808** | 3.36 → **0.343** |
| `_foreach_mul_.Tensor` | 1.94 → **0.814** | 2.04 → **0.374** |
| `_foreach_sqrt` | 1.05 → **0.484** | 1.57 → **0.302** |

Every "before" for the four `.Scalar`/`.ScalarList` cells is freshly measured;
the others were confirmed within the ±8% dead band of their recorded values.
All 16 recorded entries were re-run and reproduced. `_foreach_norm` and
`_fused_adamw_` are untouched and their baselines were not written.

## Semantics

- mul/add accept every float dtype: the per-tensor kernels they replace widen to
  fp32 and narrow the same way, so the batched result is bit-identical (asserted
  at `rtol=0, atol=0` for f16/bf16). div, lerp, addc\*, sqrt and mul.Tensor stay
  fp32 — `div_.Scalar` divides in the operand dtype, so widening would *not* be
  bit-identical.
- Anything the batched launch cannot order correctly (aliasing or duplicated
  entries, mixed dtypes, strided members, non-float lists) still returns
  `NOT_HANDLED` and reaches ATen's sequential decomposition.
- New tests: offset-view bases (4/8/12 bytes past a 16-byte boundary) with
  lengths that are not multiples of the vector step, exercising the per-chunk
  peel and the scalar tail; a staggered variant where the operands' residues
  disagree so no peel can align them and the kernel falls to its element-at-a-
  time route; the same for `sqrt`, whose output is aligned while its input is
  not; reduced-precision mul; and a mixed-dtype list that must fall back.
- `tests/`: **3117 passed**, 0 failed (41 skipped, 95 xfailed, 1 xpassed).

## Other GPUs

`scripts/compare_kernel_asm.py` vs `main` for **gfx942**: `FusedAdamW`,
`ForeachL2Norm` and `ForeachGatherScalars` device code is byte-identical, as are
the sampled `elementwise_ops` kernels; the only device-code change is the
intended replacement of the mul-tensor kernel. Note the launch geometry is host
code: it scales with the runtime CU count, but the 4-blocks-per-SM constant was
fitted on this card, so its effect on AMD is **unmeasured**. Incidentally, `main`
cannot even *build* the Metal-ABI foreach variants for gfx942 (`function
instantiation failed` on the 8-pointer launch); this PR keeps them out of the
non-Apple build entirely.

**apple-m4**: cross-compilation of these kernels fails identically on both trees
from this Linux box (pre-existing, same terminal failure as above), so the Metal
half is verified only by type checking through to the same launch call `main`
makes — its kernels are unchanged source. A non-fp32 list on Metal now falls back
instead of reaching the descriptor kernel, which on Metal would have read
addresses back as zeros.

Closes #376

### DIFF
--- a/torch_mojo_backend/eager_kernels/foreach_batched_kernels.mojo
+++ b/torch_mojo_backend/eager_kernels/foreach_batched_kernels.mojo
@@ -1,0 +1,599 @@
+"""One descriptor-batched kernel body for the whole foreach elementwise family.
+
+`aten::_foreach_*` is a list-of-tensors op: the same elementwise math applied
+to every tensor of a TensorList. Running it as one ATen call per tensor is
+dispatch bound (an optimizer step touches dozens of parameters), so the list
+has to become ONE launch, and the grid has to describe the WORK rather than
+the list.
+
+This file is that single implementation. `_foreach_ew_kernel` is
+comptime-parametrized by
+
+  * `dtype` -- any of `op_utils.FLOAT_DTYPES`; every op computes in FP32 and
+    narrows back, which is exactly what the sequential `elementwise_ops` /
+    `logic_ops` kernels these paths replace do, so the batched result stays
+    bit-compatible with the per-tensor decomposition;
+  * `op` -- which element math to apply: mul/add/div by a host scalar (one per
+    tensor, so `.Scalar` and `.ScalarList` are the same code), multiply by a
+    device-resident scalar tensor, scalar lerp, addcmul, addcdiv, sqrt.
+
+Every tensor is cut into `chunk_elements`-sized chunks and one block takes one
+chunk, with the descriptor's `chunk_end` a running prefix sum of chunk counts
+so a flat block index maps back to (tensor, element range).
+
+`chunk_elements` is a runtime launch argument, and that is the whole
+performance story of this file: a chunk fixed at 65_536 elements gives a
+four-tensor, 1M-elements-each list 64 blocks, which leaves more than half of
+an H100's 114 SMs idle for the entire launch, and gives a sixteen-tensor
+64k-elements-each list 16 blocks. Sizing the chunk from the element count and
+the SM count instead puts both on a few hundred blocks. Being a launch
+argument and not a compile-time parameter, every shape reuses one compiled
+kernel -- no shape is baked in anywhere.
+
+Apple GPUs cannot take this route at all: Metal only translates pointer-typed
+kernel *arguments* into GPU addresses, so a descriptor carrying raw addresses
+reads back as zeros (measured; see `foreach_elementwise_kernels`).
+`_foreach_ew_enqueue` therefore regroups the very same descriptors into that
+file's fixed-arity Metal launches on Apple, and takes the descriptor route
+everywhere else.
+"""
+
+from std.builtin.device_passable import DevicePassable, DeviceTypeEncoder
+from std.collections import InlineArray
+from std.gpu import block_idx, thread_idx
+from max.gpu.host import DeviceContext
+from std.math import ceildiv, min
+from std.sys.info import has_apple_gpu_accelerator, size_of
+
+from foreach_elementwise_kernels import (
+    FEA_ADDCDIV,
+    FEA_ADDCMUL,
+    FES_ADD,
+    FES_DIV,
+    FES_MUL,
+    FOREACH_EW_CHUNK,
+    FOREACH_EW_SLOTS,
+    enqueue_foreach_addc_f32,
+    enqueue_foreach_lerp_f32,
+    enqueue_foreach_mul_tensor_f32,
+    enqueue_foreach_scalar_f32,
+    enqueue_foreach_sqrt_f32,
+)
+from op_utils import _enqueue_cached, _make_ptr, _device_sm_count, ieee_sqrt
+
+
+# Element math selectors. `.Scalar` and `.ScalarList` share a code: the scalar
+# is per tensor either way.
+comptime FEW_MUL = 0  # self *= scalar[i]
+comptime FEW_ADD = 1  # self += scalar[i]
+comptime FEW_DIV = 2  # self /= scalar[i]
+comptime FEW_MUL_TENSOR = 3  # self *= *scalar_addr (a 0-d device tensor)
+comptime FEW_LERP = 4  # self = lerp(self, end, weight)
+comptime FEW_ADDCMUL = 5  # self += scalar[i] * (t1 * t2)
+comptime FEW_ADDCDIV = 6  # self += scalar[i] * (t1 / t2)
+comptime FEW_SQRT = 7  # out = sqrt(in), the one out-of-place member
+
+comptime FEW_THREADS = 256
+
+# Blocks per SM the chunk size aims for. Swept on an H100 PCIe (114 SMs, FP32),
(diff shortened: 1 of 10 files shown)


gabrieldemarmiesse:
Now that we have the numbers for the benchmark for apple, can you update them too?

torch_mojo_backend/eager_kernels/foreach_batched_kernels.mojo
```diff
@@ -0,0 +1,599 @@
...
+                slot += 1
+            if group_chunks == 0:
+                continue
+            _few_backfill(first_addrs, used_slots)
+            comptime if _few_addrs[op]() >= 2:
+                _few_backfill(second_addrs, used_slots)
+            comptime if _few_addrs[op]() >= 3:
+                _few_backfill(third_addrs, used_slots)
+
+            comptime if op == FEW_MUL:
```
gabrieldemarmiesse: can't this big list of if-elif-elif... be a comptime for + a comptime if instead with a check at the end to make sure we launched a kernel?

gabrieldemarmiesse (changes requested):

torch_mojo_backend/eager_kernels/foreach_batched_kernels.mojo
gabrieldemarmiesse-ai-agent: Both done in `16c0cd0`, and the guard caught a real bug rather than just tidying the shape.

**The `else` was launching sqrt.** The chain ended in a bare `else:` calling `enqueue_foreach_sqrt_f32`, so on Apple *any* foreach op without an explicit arm would have silently computed a square root of its operands and returned successfully — wrong numbers, no error. `sqrt` is now an explicit `comptime if`, which is what removes it.

**Five of seven arms collapse**, into two generated bodies, because they reach the same entry point and differ only in a comptime code:

- `mul` / `add` / `div` -> `enqueue_foreach_scalar_f32[FES_*]`, driven by `_FEW_SCALAR_FAMILY`
- `addcmul` / `addcdiv` -> `enqueue_foreach_addc_f32[FEA_*]`, driven by `_FEW_ADDC_FAMILY`

`mul.Tensor`, `lerp` and `sqrt` have genuinely different argument lists and stay explicit. Forcing those into the same loop would have traded readable duplication for unreadable cleverness, so I left them.

**The check you asked for is compile-time.** `foreach_ew_enqueue` opens with `comptime assert _few_metal_covered[op]()`, and two details make that trustworthy rather than decorative:

- the predicate is **computed from the same three rosters the arms are generated from** (`_FEW_SCALAR_FAMILY`, `_FEW_ADDC_FAMILY`, `_FEW_SOLO`), so it cannot drift from them — which is the usual way a hand-maintained "supported ops" list rots;
- it lives in the **shared** entry point, not the Apple branch, so a missing arm fails the build on any host rather than only on a Mac nobody runs.

Verified negatively, which is the only way that claim means anything: deleting `FEW_SQRT` from the roster and building `-D OP=ForeachSqrt` for **sm_90a** fails with `constraint failed: this foreach op has no arm in _foreach_ew_enqueue_apple`. Behind it a folded-away `if not launched: raise` catches the opposite drift (listed as covered, arm deleted); every assignment to `launched` sits in a `comptime if`, so it constant-folds out for a covered op.

**Three more silent fallthroughs in the same file** got the same treatment, each now ending in `comptime assert False`: `_few_addrs` (would mispack metadata with a plausible arity), `_few_label` (would print **another op's algorithm name into a user's profiler**) and `_few_element` (would apply addcdiv's arithmetic).

**Invariance: sm_90a 0 of 16 kernels differ, apple-m4 0 of 16 differ**, 0 build errors either side, across all 8 ops x {f32, bf16}. Note the second number — because the `InlineArray[Int64, 8]` fix landed, Metal now cross-compiles from this box, so the Apple half is genuinely device-code-verified rather than type-checked only.

Worth recording for anyone touching this family: `compare_kernel_asm.py` gives a **false all-clear** here. The family registers through a non-literal gate name so the script's own scan cannot enumerate its ops, and `--ops` only filters that scan. The ops have to be forced by name. Concurrent runs sharing one work directory also produced a bogus "before 2, after 0".

`tests/`: **3133 passed, 0 failed.** No baseline re-recorded — this is behaviour-neutral apart from the removed bug.

PR #380, opened 2026-08-14
Record the Apple foreach baselines, and fix the Metal build that blocked them

gabrieldemarmiesse-ai-agent:
Follow-up to #379, asked for on that PR: record the Apple foreach benchmark
numbers now that the M4 has a baseline block.

**They were never recorded because the kernels did not build there.** #377 gave
the M4 716 leaves and zero `_foreach_*` entries (only the two `_fused_adamw_`
ones). Running `benchmarks/test_foreach.py` on the Mac fails 18 of its 20 nodes
with an `ImportError` from `mojo build`, on `main` and on #379 alike:

```
constraint failed: Int and UInt do not conform to DevicePassable;
use a fixed-width type such as Int32 or Int64 instead
```

**#379 did not fix it, so this PR does.** The Metal half of the family
(`foreach_elementwise_kernels.mojo` — the fixed-arity kernels Metal needs
because it only translates pointer-typed *arguments* — plus the two batched norm
kernels in `foreach_clip_kernels.mojo`) passes its per-slot chunk offsets and
lengths as `InlineArray[Int, 8]` kernel arguments. `InlineArray` encodes
element-wise, so the *element* type has to be fixed width too, and the mojo-1.0
migration that renamed lone scalar parameters to `x_arg: Int64` never covered
the arrays. They now cross the ABI as `InlineArray[Int64, 8]` (`_SlotInts`),
widened at the enqueue boundary by `_slot_ints` so the host-side regrouping
keeps working in `Int`; kernels convert back at the use site and every index
computation stays in `Int`.

Nothing here is reachable off Metal — the CUDA/ROCm route is the
descriptor-batched body in `foreach_batched_kernels.mojo`, selected by
`comptime if has_apple_gpu_accelerator()` — which is why only a Mac ever saw
this, and why the fix cannot move another architecture.

**sm_90a is byte-identical**: 0 of 22 kernels differ across every `Foreach*`
specialization of `optimizer_ops` in float32 and bfloat16, plus 0 of 16
sidecars for the three ops the scanner finds on its own. Note for whoever runs
this next: `scripts/compare_kernel_asm.py` alone reports a **false all-clear**
here — `optimizer_ops` registers the family through `_register_foreach_ew[op,
name]`, a non-literal gate name, so the source scan marks the module opaque and
builds only `FusedAdamW`, `ForeachL2Norm` and `ForeachGatherScalars`. The eight
`Foreach*` ops have to be forced by name. Host launch geometry is unchanged too:
the only edits outside the Apple-gated branch are the two type names.

## Apple foreach coverage — 17 recorded, 1 refused, 2 already there

| node | L_4x1048576 | L_16x65536 |
| --- | --- | --- |
| `_foreach_add_.Scalar` | 0.918 | 0.773 |
| `_foreach_addcdiv_.ScalarList` | 1.057 | 0.624 |
| `_foreach_addcmul_.Scalar` | 1.055 | 0.635 |
| `_foreach_div_.ScalarList` | 0.927 | 0.802 |
| `_foreach_lerp_.Scalar` | 1.065 | 0.803 |
| `_foreach_mul_.Scalar` | 0.921 | 0.787 |
| `_foreach_mul_.Tensor` | 0.924 | 0.999 |
| `_foreach_norm.Scalar` | 0.843 | 0.587 |
| `_foreach_sqrt` | 1.503 | **refused** |

(ratio ours/stock MPS device time; below 1.0 means we are faster.)

`_foreach_sqrt/f32/L_16x65536` was refused as *too noisy to trust* on two
separate runs — ~150 µs of work, leg spread 67% then 84%, median uncertainty
still above 1.5% after the 12-burst cap on all 3 internal attempts. It stays
unrecorded rather than recorded wrong, so the M4 foreach coverage is **17 of
18** previously-missing nodes. The two `_fused_adamw_` entries #377 already had
still pass unchanged.

Apple config goes 716 → 733 leaves; the H100 block is unchanged key-for-key and
value-for-value, and the hand-written viewer half of `baselines.html` is
byte-identical. `benchmarks/report.py` reads the file and
`benchmarks/test_coverage.py` passes.

## Layering

This branch is **#379's head plus one commit** — the Metal fix is a fix to
#379's files, and the numbers were measured with #379's code — so it should land
**after #379**. #377 needs no special handling: it is already merged into `main`
(58a486c) and #379 has merged `main`, so the file here carries both Apple sets
and this PR only adds to them.

## Unrelated pre-existing failure worth knowing about

On the Mac, `tests/test_eager_optimizer_ops.py` is 50 passed / 2 failed with
this branch, and the same 2 fail identically on unpatched #379 (same 4502
mismatched elements), so they are #379's, not this PR's:
`test_batched_foreach_mul_scalar_reduced_precision[f16]` and `[bf16]`. The Apple
batched path is float32-only, so half-precision lists fall back to a per-tensor
multiply that computes in the narrow dtype and lands one ulp off ATen, which
widens to fp32 and narrows once. Our own `mul_.Scalar` on Metal *is* bit-exact
against the widened reference (checked directly: 0 of 65539 elements differ), so
the gap is in the fallback route the f16 list takes, not in the elementwise
kernel. Benchmarks are f32 only, so the numbers above are unaffected.

### DIFF
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -723,6 +723,163 @@ document.addEventListener("DOMContentLoaded", boot);
             }
           }
         },
+        "_foreach_add_.Scalar": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.773
+                  }
+                },
+                "L_4x1048576": {
+                  "layouts": {
+                    "contig": 0.918
+                  }
+                }
+              }
+            }
+          }
+        },
+        "_foreach_addcdiv_.ScalarList": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.624
+                  }
+                },
+                "L_4x1048576": {
+                  "layouts": {
+                    "contig": 1.057
+                  }
+                }
+              }
+            }
+          }
+        },
+        "_foreach_addcmul_.Scalar": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.635
+                  }
+                },
+                "L_4x1048576": {
+                  "layouts": {
+                    "contig": 1.055
+                  }
+                }
+              }
+            }
+          }
+        },
+        "_foreach_div_.ScalarList": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.802
+                  }
+                },
+                "L_4x1048576": {
+                  "layouts": {
+                    "contig": 0.927
+                  }
+                }
+              }
+            }
+          }
+        },
+        "_foreach_lerp_.Scalar": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.803
+                  }
+                },
+                "L_4x1048576": {
+                  "layouts": {
+                    "contig": 1.065
+                  }
+                }
+              }
+            }
+          }
+        },
+        "_foreach_mul_.Scalar": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.787
+                  }
+                },
+                "L_4x1048576": {
+                  "layouts": {
+                    "contig": 0.921
+                  }
+                }
+              }
+            }
+          }
+        },
+        "_foreach_mul_.Tensor": {
+          "dtypes": {
+            "f32": {
+              "shapes": {
+                "L_16x65536": {
+                  "layouts": {
+                    "contig": 0.999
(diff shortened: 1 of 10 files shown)


gabrieldemarmiesse:
Can you close this and put this directly in #379?

gabrieldemarmiesse-ai-agent:
Closing — folded directly into #379 as you asked. Its single commit `2ef17b4` ("Make the Apple foreach kernels build again, and record their baselines") is now the head of #379, applied as a clean fast-forward so nothing was rewritten and no history was lost.

#379 now carries all three commits:

```
b83abb9  One batched kernel body for the whole foreach elementwise family
064a58f  Merge branch 'main' into improve-foreach-mul-scalar
2ef17b4  Make the Apple foreach kernels build again, and record their baselines
```

Everything from this PR's description applies there instead: the `InlineArray[Int, 8]` -> `InlineArray[Int64, 8]` fix for the Metal build, sm_90a verified 0 of 22 kernels changed, and the 17 Apple foreach baselines. Verified after the move: Apple 716 -> 733 leaves with 0 changed, H100 857 -> 857 with 0 changed, and the viewer half of `baselines.html` byte-identical.

PR #387, opened 2026-08-16
Route dense conv through tensor-core GEMM + sm_90a implicit-GEMM (TMA im2col)

gabrieldemarmiesse-ai-agent:
## Is this win real? (honest-reference audit)

A maintainer question deserves a direct answer before anything else in this
description: **partially.** The headline number this PR has carried
(`0.65x cuDNN`) is correct but was computed against a reference that is not
what "cuDNN" means to most readers. Restated against both references:

**0.65x torch's default NCHW conv path; 1.15x cuDNN at equal layout
(channels_last).**

Both numbers are true and both matter, because they answer different
questions. The 0.65x number is what a user actually sees when they call
`F.conv2d` on an ordinary (NCHW) tensor, which is what nearly everyone does
-- so it is not a fake number. But it flatters this route, because cuDNN's
own convolution kernels are NHWC-only: reached from NCHW, torch pays two
layout-transpose kernels around cuDNN's actual (fast) convolution, and this
route needs no such transpose. Handed the layout cuDNN actually wants, the
comparison flips.

**Body-shape decomposition** (`N32xC64x56x56_K64k3s1`, bf16, device time, us):

| leg | total | breakdown |
|---|---:|---|
| stock, NCHW input (what `F.conv2d` does by default) | 101.6 | 31.2 conv + 41.0 nchw&harr;nhwc transposes + 29.9 bias |
| stock, channels_last input (the layout cuDNN wants) | 58.8 | conv + bias, no transposes |
| **ours** | **66.5** | NHWC pass + weight repack + implicit GEMM + bias |

`66.5 / 101.6 = 0.65` (the recorded number). `66.5 / 58.8 = 1.13-1.15`
depending on which pair of runs is compared (the source comment in
`aten_fast.py` now states both legs instead of only the favorable one).

**Every recorded node, both references** (bf16, device time, ours/stock,
lower is better):

| shape | route | vs stock NCHW (recorded) | vs stock channels_last (fair) |
|---|---|---:|---:|
| body `N32xC64x56x56_K64k3s1` | implicit GEMM | 0.653 | **1.147** |
| stem `N8xC3x224x224_K64k7s2` | round 1 | 1.074 | **1.455** |
| mid `N32xC256x14x14_K256k3s1` | implicit GEMM | 1.452 | **2.402** |
| awkward `N7xC48x39x53_K80k3s1` | implicit GEMM | 1.102 | **1.929** |
| pw1x1 `N32xC256x56x56_K64k1s1` | round 1 (unchanged) | 0.783 | **0.846** |

At the fair reference, four of the five nodes this PR reported as wins are
losses -- including the flagship body shape, whose "0.65x cuDNN" becomes a
1.15x loss once cuDNN is not forced to pay a transpose tax it doesn't
actually need. **`pw1x1` is the one genuine win**, and it predates this PR
(round 1, unaffected by the implicit-GEMM work) -- a 1x1 conv has no im2col
on either side, so there is no transpose asymmetry to exploit either way.

This does not mean the kernel is bad, or that the design is wrong: putting
`out_c` on the GEMM's M axis so the output lands in NCHW without a
transpose (the design's actual contribution) is a real, correct idea, and
against the reference this repo's own benchmark suite has always used for
every other op (`F.conv2d` on an NCHW tensor, i.e. torch's default), the
0.65x/1.07x/0.78x numbers on body/stem/pw1x1 are genuinely what a default
caller sees. It means the *margin* over cuDNN was overstated: this route is
currently ~1.1-2.4x slower than cuDNN at equal layout, and closing that
gap -- not layout arbitrage -- is the actual remaining work. See "What's
next" below.

**The one-time cost was not free either.** `torch.nn.functional.conv2d` and
this route both pay one layout conversion (torch's two transposes on the
NCHW leg above; this route's own NHWC pass), so the 0.65x number is not
"comparing our transpose cost to none" -- it is comparing our ~1 transpose
of one operand to cuDNN's ~2 transposes of input and output. That
asymmetry is real and is exactly what round 2 exploited; it is also exactly
why it stops being a win once cuDNN's transposes are removed from its own
side of the comparison instead.

### New coverage: what happens outside the implicit-GEMM's domain

The five nodes above are all inside the implicit-GEMM's domain (dense,
stride 1, dilation 1). `benchmarks/test_vision.py`'s `CONV_SHAPES` gained 8
shapes chosen to sit outside it, so the recorded suite describes the whole
op rather than one fast path (device time, ours/stock vs NCHW, both dtypes,
plus bf16 against the fair channels_last reference -- that column is the
audit's own measurement, not yet a recorded baseline):

| shape | regime | bf16 (suite ref, NCHW) | f32 (suite ref, NCHW) | bf16 (fair ref, channels_last, audit measurement) |
|---|---|---:|---:|---:|
| `N32xC128x28x28_K128k3s2` | stride 2 | 12.566 | 12.283 | 23.1 |
| `N16xC96x28x28_K192k3s2` | stride 2, ragged C/K | 5.337 | 7.757 | 9.74 |
| `N32xC256x14x14_K512k1s2` | 1x1, stride 2 (ResNet shortcut) | 3.482 | 4.427 | 6.23 |
| `N16xC64x32x32_K64k3s1d2` | dilation 2 | 6.791 | 7.961 | 12.86 |
| `N1xC64x56x56_K64k3s1` | batch 1 | 1.375 | 4.026 | 2.43 |
| `N1xC256x14x14_K256k3s1` | batch 1, deeper C | 1.806 | 2.278 | 3.30 |
| `N32xC512x7x7_K512k3s1` | C=K=512, weight-repack-bound | 2.019 | 18.407 | 4.10 |
| `N8xC128x112x112_K128k3s1` | large spatial (VGG-style) | 0.981 | 7.679 | 1.81 |

Stride 2 and dilation are declined by the implicit GEMM by design (its box
convention only produces the exact output grid at stride 1) and fall back
unchanged to round 1's materialized-im2col route, which was never tuned for
them: **3.5x-12.6x slower than stock NCHW**, and worse still at the fair
channels_last reference (**1.8x-23.1x** across this whole set, not just the
stride/dilated rows -- the layout-arbitrage effect that flatters the five
original nodes flatters these just as much, so their fair-reference numbers
are uniformly higher than their suite-ref ones). The mechanism for the
stride/dilation rows is a non-coalesced gather: the vectorized `Im2col`
kernel's fast path reads a contiguous run of input columns when
`stride_w == 1`, but at stride 2 every other element is skipped, and the
measured achieved bandwidth drops to roughly 40 GB/s against the ~600 GB/s
the coalesced stride-1 path reaches on the same kernel -- the gather, not
the GEMM, is why these shapes are slow. `N32xC512x7x7_K512k3s1` is a
different failure mode entirely: it *is* inside the implicit-GEMM's domain,
and its bf16 number (2.019 suite-ref, 4.10 fair-ref) is unremarkable, but
f32 (18.407, like every f32 number here) never reaches
that route at all -- see the TF32 note below -- and separately, weight
repack (`(out_c, K_pad)`, here 512 x 4608 elements) becomes a
larger fraction of the total at this C/K, which is one of the follow-ups
below regardless of dtype.

**A channels_last-native caller pays extra, not less.** A model already
laid out `channels_last` (a standard PyTorch AMP/inference optimization) is
exactly the case where cuDNN needs no transpose, so it is the case where
this route currently loses most -- and it loses by more than the numbers
above: this route's own preflight (`_tc()`) materializes ANY non-contiguous
input back to dense NCHW before its kernels run, so a channels_last caller
first pays a permute-to-NCHW this route requires, then the 1.1-2.4x it
costs versus NCHW-native cuDNN, then would need to permute back if the rest
of their model expects channels_last. Measured on the body shape: **2.7x
worse than the NCHW-input number** once that forced permute is counted.

**The f32 numbers are not comparing the same math.** Stock PyTorch's cuDNN
convolution defaults `torch.backends.cudnn.allow_tf32 = True` -- f32 conv
runs on TF32 tensor cores unless the user opts out. This route's f32 gate
(`_try_tf32_conv_bmm`) instead follows `torch.get_float32_matmul_precision()`
(default `"highest"`, TF32 off), the same policy every other TF32 gate in
this repo uses for mm/bmm. At default settings, then, every f32 number
above is **true fp32 compute compared against stock's TF32 compute** -- a
real precision difference, not just a speed one. This was disclosed in
round 1's numerics section and remains true here; it is restated because it
is the single largest reason the f32 column looks as bad as it does, and
conflating it with "the kernel is slow" would be its own honesty problem.
See the open question below.

**Correctness was not part of what the audit found wrong.** A 624-case
edge sweep across shapes, dtypes, stride, dilation, groups and both routes
found 0 wrong numbers and 0 unexpected dispatch declines (every case that
should reach a given route did, every case that should decline did), and
the timing decomposition above was independently re-verified to sum
correctly against wall-clock totals on each leg. What was wrong was the
choice of reference and the framing built on it, not the kernel's answers
or its dispatch.

## Diagnosis (round 1)

`fast_aten_convolution` lowered every dense conv to im2col + `_MatmulExtension`
"Bmm"/"Matmul" -- the generic SIMT GEMM in `matmul_ops.mojo`: fp32 FMA compute
on NVIDIA regardless of dtype (the file's only `mma()` use is AMD
gfx942-gated). Meanwhile bf16/f16 mm/bmm/linear/attention already route
through the H100 tensor-core `Gemm16`/`Bmm16` kernels, and f32 mm/bmm already
has an opt-in TF32 route -- conv used neither.

Profiling the body benchmark shape (`N32xC64x56x56_K64k3s1`, bf16) after
wiring the GEMM alone showed the GEMM itself was only ~9% of device time --
Im2col was 87%. `ncu` traced that further to two things: `sm__throughput` at
70-78% against `dram__throughput` at 2-5% (compute-bound, not memory-bound),
and the specific compute was the per-*element* `(s, c, fh, fw, oh, ih)`
index decomposition the generic one-thread-per-output-element `elementwise`
kernel redid on every one of `batch*channels*kh*kw*out_h*out_w` threads.
`BiasAddChan` (the epilogue add) had the same shape of problem at smaller
scale.

## Round 1: tensor-core GEMM routing + vectorized im2col/bias-add

Three independent fixes, each falling back cleanly when it doesn't apply:

**1. Tensor-core GEMM routing** (`aten_fast.py`, Python only, zero `.mojo`
changed). Dense (`groups == 1`) conv's GEMM tries, in order: `_try_gemm16_conv_bmm`
(bf16/f16 on sm_90a, through the *exact* `Bmm16` kernel `fast_aten_bmm`
already builds -- conv's shared-weight batched GEMM exploits the same
broadcast-A mechanism `Bmm` already threads through), `_try_tf32_conv_bmm`
(f32 on sm_90a, opt-in, gated like `_try_tf32_mm`/`_try_tf32_bmm`), then the
untouched SIMT `Bmm` fallback.

**2. Vectorized Im2col** (`conv_ops.mojo`, GPU only). One warp per `(sample,
channel, kh-tap, kw-tap, out-row)`, lane 0 computes the index decomposition
once and `shuffle_idx` broadcasts it -- no shared memory, no barrier.

**3. Vectorized BiasAddChan** (`conv_ops.mojo`, GPU only). Same one-warp-per-row
restructuring for the epilogue add.

Numerics: PyTorch's cuDNN conv defaults `allow_tf32 = True` but this repo's
`_try_tf32_conv_bmm` follows `torch.get_float32_matmul_precision()` (default
`"highest"`, TF32 off) like every other TF32 gate here -- so at default
settings f32 conv's GEMM is on the untouched SIMT path; f32's round-1 gain
below is Im2col/BiasAddChan vectorization only. Known, intentional asymmetry,
not a bug (and the same asymmetry the audit above re-flags for round 2's
numbers).

## Round 2: sm_90a implicit-GEMM (TMA im2col, no materialized buffer)

Round 1 made the GEMM fast but left the *materialized im2col* itself as the
new bottleneck: on the body shape it was 87% of device time before round 1,
and after round 1 the GEMM alone still cost ~140us against a ~101us *total*
stock (NCHW) reference -- profiling showed the materialized im2col write+read
is not actually a memory-traffic wall (150 GB/s against ~2 TB/s HBM, i.e.
7%) but a store-transaction-bound one, and even a perfect-bandwidth im2col
would still total ~2.8x that reference.

The fix: Hopper's hardware im2col TMA
(`cp.async.bulk.tensor.4d.shared::cluster.global.im2col`) reads each GEMM B
tile straight out of an NHWC copy of the activation -- no patch matrix is
ever written to memory. Putting the weight (not the activation) on the GEMM's
M axis makes the output tile `(out_c x pixels)`, i.e. already NCHW, so this
route needs no output transpose. Only one NCHW->NHWC pass on the *input*
remains (TMA requires the descriptor's innermost dim contiguous, and for
im2col that's C). As the audit above shows, this NHWC pass is where the win
over an NCHW-input reference comes from, and also exactly why the win
shrinks or reverses once the reference is not paying for its own layout
conversion.

The mainloop/epilogue reuse the batched-BMM `Bmm16` body
(`_b5_bmm_nn_tiny`) verbatim via a new shared `wgmma_c_epilogue.mojo` --
only the B producer (the im2col TMA issue) is new code
(`conv_igemm_ops/conv_igemm_ops.mojo`).

Domain (checked at the call site, before any allocation): `dtype ==
bfloat16` (f16/f32/TF32 not built -- see "What's next"), `sm_90a`, `groups ==
1`, `stride == dilation == 1` (the box convention this kernel uses is only
valid there), `R*S > 1` (a 1x1 conv has no im2col at all on either route and
was already fast), `C >= 32` and `out_c >= 32` (a thin-C stem would pay a
21x-inflated reduction from the 64-channel pad; a thin `out_c` wastes half
of every 64-row output tile), and a host-side TMA descriptor legality check
(pixel-box corners in `[-128,127]`, filter taps in `[0,255]` -- both are
hardware limits of `cuTensorMapEncodeIm2col`/the im2col instruction, and an
illegal tap is undefined behaviour rather than a loud failure, so the gate
exists specifically to keep that case off the route). Everything the gate
declines, grouped convs and stride/dilation included, falls through
unchanged to round 1's tensor-core-GEMM route -- unchanged means unchanged:
those regimes are not slower because of this PR, but they are also not any
faster, and the new coverage above puts numbers on that for the first time.

Developed and adversarially reviewed (via a from-scratch Codex pass: 3
MUST-FIX + 4 SHOULD-FIX) in a standalone harness before integration. All 7
findings are applied in the integrated kernel: no guard-sample
over-allocation (an earlier design over-allocated the activation buffer by
one full sample to dodge an OOB-fault edge case that turns out to be
unnecessary -- the hardware already zero-fills that tail), the descriptor
domain gate above (with the review's own adversarial case -- a shape that
passes every other eligibility rule but whose last filter tap is illegal --
as a permanent unit test), compile-time tile-legality asserts (an illegal
`(bn, ppc)` pairing used to be a silent runtime hang, now a build error),
explicit 8-byte shared-memory alignment (Mojo defaults to align-4 and
silently scalarizes vector accesses), edge-case tests that check every
output element for rectangular filters / asymmetric padding / batch=1
(sampling only every Nth element, as an earlier version did, cannot see a
tail-only indexing bug), a tolerance derived from the fp32 accumulation
error bound (`eps_f32 * sqrt(K) * sum|a*b|`) plus a bit-exact-fraction
assertion instead of a fixed absolute+relative slop, and launch-bounds
metadata on every kernel.

### Why `mid` and `awkward` are the furthest from closing the gap

Both are limited by `BM = 64`: the activation tile is re-read `out_c / 64`
times (4x on `mid`), so `mid`'s GEMM alone moves ~226 MB through L2 in ~46us
= 4.3 TB/s, which *is* L2 peak on this card. The fix is a `BM = 128` variant
(two consumer warp groups) that halves that re-read traffic -- nothing about
the im2col producer needs to change. This is the same batched-GEMM-tuning
gap round 1 flagged for the pre-round-2 route; it is inherited here because
the mainloop is shared, not because implicit GEMM introduced a new
limitation. It is also, per the audit above, not enough by itself to reach
parity with cuDNN's own convolution -- it closes the GEMM-side gap, not the
1.1-2.4x fair-reference gap as a whole.

## What's next

Reordered from the audit's findings, highest expected value first:

1. **Stride-2 / dilated Im2col.** The single biggest number in this PR is a
   regression, not a win: 3.5x-12.6x slower than stock on shapes outside the
   implicit GEMM's domain, from a non-coalesced gather in the *unmodified*
   round-1 kernel (~40 GB/s achieved against ~600 GB/s coalesced on the
   same kernel at stride 1). Fixing the gather pattern for `stride_w != 1`
   (vectorize per-group instead of per-element, or reroute through a
   stride-aware TMA descriptor) does not require touching the implicit-GEMM
   kernel at all and looks like the highest-value remaining work.
2. **Weight-repack cost at large C/K.** `N32xC512x7x7_K512k3s1` shows the
   `(out_c, K_pad)` repack becoming a larger share of total time as C and
   `out_c` grow; vectorizing it (8 contiguous k-elements/thread, ~6.6us on
   the body today) matters more at these shapes than it did on the original
   benchmark set.
3. **`channels_last`-native support.** This route's `_tc()` preflight
   forces every input to dense NCHW; a caller already in `channels_last`
   pays for an unnecessary permute in and (if the rest of their model
   expects it) another one out. Accepting the activation directly in NHWC
   would remove both the audit's "forced permute" penalty and this route's
   own remaining NCHW->NHWC pass in the same change.
4. **`BM = 128` batched GEMM body.** Still worth doing (see above), but
   reordered behind the first two: it closes the GEMM-side gap on `mid`/
   `awkward`, which the audit shows is a smaller piece of the fair-reference
   gap than stride/dilation coverage or channels_last support.
5. **f32 / TF32 implicit GEMM**, and **`stride != 1`/dilation support
   directly in the TMA descriptor** (two routes identified: a local copy of
   `create_tensormap_im2col` with `elementStrides` set, or a
   `pixels_per_column = 1` halo-box variant with absolute coordinates),
   remain open and are partly redundant with item 1 if the gather fix
   above is done at the TMA level instead of in the fallback kernel.
6. **Grouped convolution** stays on round 1's route; not attempted for
   implicit GEMM here.

## Verification

- **Correctness**: `tests/test_eager_kernels.py` -- 33 new implicit-GEMM
  tests (12 shapes x with/without bias, covering both TMA-store and scalar
  epilogues, rectangular filters, asymmetric padding, batch=1, C-padding,
  ragged `out_c`; 8 domain-decline cases -- stride/dilation/grouped/1x1/thin-C/
  thin-out_c/f32/f16 -- asserting they still compute correctly on the
  round-1 route; the descriptor-domain gate unit test) plus the pre-existing
  63 conv tests and 140 gemm16/bmm tests, all pass. The audit's own 624-case
  edge sweep (above) is a separate, broader correctness pass across both
  routes: 0 wrong numbers, 0 unexpected declines.
- **Blast radius**: the implicit-GEMM kernel is a new module
  (`conv_igemm_ops`), zero risk to anything that doesn't call it. The one
  shared-code change (`gemm16_bmm_v5_kernels.mojo`'s C-tile epilogue
  extracted into `wgmma_c_epilogue.mojo` so the conv kernel reuses it
  instead of duplicating ~60 lines) touches the production `Bmm16` kernel
  used by every bf16/f16 mm/bmm/linear/attention call on sm_90a.
  `scripts/compare_kernel_asm.py --accelerator sm_90a --modules
  gemm16_matmul_ops` shows 6 of 344 built kernels differ -- all in the
  scalar (non-TMA-store) epilogue of the NN-layout batched body. The diff
  is NOT byte-identical: agent-C's independent pass confirmed it as
  register renumbering plus rescheduled parameter loads and one rescheduled
  multiply -- a new-function-boundary artifact of `@always_inline`, not a
  changed operation -- and traced the dispatcher
  (`gemm16_v3_kernels.enqueue_gemm16_bmm`) to confirm only NN-layout
  bf16/f16 batched matmul reaches the touched body at all (TT/NT/TN layouts
  and all tf32 traffic go through entirely different kernel families).
  `benchmarks/test_gemm.py -k bmm` showed 9 unrelated regressions, all
  TT-layout or tf32 -- matches this repo's documented GEMM-baseline
  best-of-N flakiness, not this change. On gfx942, the new module's main
  kernel cross-compiles to a 0-vgpr/0-sgpr stub (the `comptime if
  _is_sm_9x()` gate elides the body entirely off Hopper), and the
  Python-level dispatch additionally never calls into this extension off
  `sm_90a`, so AMD is double-gated. `benchmarks/test_coverage.py`: clean.
- `benchmarks/test_vision.py -k "not conv2d"` (pooling/upsample, untouched
  by this PR): 16/16 pass, no regression.
- Baselines recorded with scoped `pytest benchmarks/test_vision.py -k
  conv2d --update-baselines` runs; every write is additive or a genuine
  improvement in an existing entry, never silently overwriting a
  regression. `benchmarks/baselines.html`'s total diff across every round
  of this branch is 114 insertions / 4 deletions (34+/4- through the
  keepalive-fix round; +80/-0 for the audit's 16 new shape x dtype leaves).

## Agent-C / agent-C2 review (independent, post-round-2)

Two independent review passes reproduced every performance and blast-radius
claim in this PR as it stood at the time (body/stem/pw1x1/awkward/mid
within 0.5% of the recorded numbers, the six-kernel scalar-epilogue diff,
the gfx942 stub) and found:

1. **MUST-FIX (agent-C) -- use-after-free under the default kernel-call
   queue.** `_try_gemm16_conv_bmm`/`_try_tf32_conv_bmm` and the plain SIMT
   `Bmm`/grouped `Matmul` fallbacks took a raw `col_ptr: int` instead of the
   tensor that owns the im2col buffer (or, on the 1x1 fast path, the input
   tensor read in place), and their `keepalive` tuples were `(out, w)` --
   missing it. Production enables the kernel-call queue by default (pytest
   disables it, which is why this passed every test at the time); under a
   cold start, the im2col item can launch and drop its own keepalive while
   the GEMM specialization is still compiling, and with nothing else
   retaining the buffer it frees before the deferred GEMM launch reads it.
   Fixed by threading the owning tensor through every one of these call
   sites instead of a bare pointer. Four new tests in
   `tests/test_call_queue.py` force the affected specializations cold,
   launch the producer alone, and check via a `weakref` (not a value
   comparison -- freed-memory reuse does not reliably corrupt a value
   promptly, which is exactly why the first attempt at this test passed on
   both the buggy and fixed code) that the buffer survives until the GEMM
   item itself launches; all four reliably fail on the pre-fix code and
   pass post-fix (verified both ways).
2. **MUST-FIX (agent-C) -- the igemm edge-shape test's comment claimed the
   harness's fp32-accumulation-derived tolerance and bit-exact-fraction
   assertion; the code used a fixed 8% slop.** Ported the actual criterion
   (`2^-8 * |acc| + 4 * eps_f32 * sqrt(k_pad) * sum|a*b|`, extended with a
   second half-ulp term for the bias-add's own independent rounding) plus
   the bit-exact-fraction check against the double-rounded reference. All
   33 igemm nodes still pass (the two deepest reductions land at ~99.87%
   exact against a 0.99 bar with `torch.randn` inputs, vs. the harness's
   hash-tuned 1.0 against its 0.999 bar -- expected, not a regression).
3. Cross-device follow-up (agent-C, not a merge blocker): `fast_aten_
   convolution`'s dispatch gate checked dtype/shape but not device equality
   between the input and the weight (or the bias); both now return
   `NOT_HANDLED` on a cross-device pair.
4. **Preflight-ordering nit (agent-C2):** the device-equality check above
   ran after `_tc()` had already materialized a non-contiguous operand
   (an allocation + copy launch), so "reject before any allocation" was not
   literally true for a non-contiguous cross-device pair (the eventual conv
   launch itself was never unsafe). Reordered to preflight dtype/device/
   shape on the raw, unmaterialized tensors and only `_tc()` the accepted
   operands afterward.

Agent-C2's verdict on the fixes above was **merge-ready**; the audit in
this description is a separate, later pass specifically checking whether
the *performance claims* -- not the correctness or safety of the code --
held up against a fair reference.

## Fixed: conv_ops.mojo did not build on a host with no accelerator

CI on this branch (run 32249078749) failed the "not mojo_device" shards
with `ImportError: mojo build failed for conv_ops`, reproducing on every
conv test including CPU-device ones
(`test_mojo_device.py::test_convolution_2d[cpu]`,
`test_high_level_ops.py::test_conv2d_with_bias[mojo:cpu, eager]`) because
the failure was at *import* time, before any device is chosen. Root cause:
`_im2col`/`_bias_add_chan` picked between the CPU scalar path and the GPU
kernel path with a runtime `if ctx.api() == "cpu": ... else: ...` -- Mojo
instantiates and type-checks BOTH branches of a runtime `if`, so the GPU
branch (`_im2col_gpu`/`_bias_add_chan_gpu`, both calling `_enqueue_cached`,
which needs a target GPU architecture at compile time) was compiled
unconditionally, even on a host with no accelerator at all. There it hits
`constraint failed: Unknown GPU architecture detected` and the whole
module fails to import. This is exactly the audience AGENTS.md's eager
rule 1 exists for -- a torch-cpu-only install must still be able to
`.to("mojo")` and use the GPU where one exists, which requires the module
to import at all on a machine that has none.

Fixed by matching the pattern already load-bearing across every other
`eager_kernels` family with a CPU/GPU split (`reduction_ops.mojo`,
`elementwise_ops.mojo`, `data_movement_ops.mojo`, `matmul_ops.mojo`'s MFMA
gate, ...): the CPU branch returns early, and the GPU branch is wrapped in
`comptime if has_accelerator():` so it is never instantiated when there is
no accelerator to target -- `ctx.api() == "cpu"` has already returned
above in that case, so the gate is provably dead code there, not a
behavior change. Verified with `scripts/compare_kernel_asm.py` (sm_90a and
gfx942, both `BiasAddChan` and `Im2col`, both dtypes): 0 kernels changed,
confirming this is host/build-structure only and moved no device code.
The full conv/igemm/queue-keepalive test selection still passes on this
(GPU-equipped) machine; the actual no-accelerator build is verified by
this push's CI run, which runs on a plain `ubuntu-latest` runner with no
GPU at all.

**Note for #418 (conv backward, conv1d):** that PR's `_col2im_gpu` will
need the identical `comptime if has_accelerator():` treatment once it
lands, for the same reason -- a runtime `ctx.api()` branch does not stop
Mojo from instantiating the GPU side on a host with none.

## CI verdict on the no-accelerator fix

Pushed (c8dab40) and confirmed: all 27 previously-red conv nodes on the
"not mojo_device" shards pass now, including
`test_mojo_device.py::test_convolution_2d[cpu]` and
`test_high_level_ops.py::test_conv2d_with_bias[mojo:cpu, eager]`. Two
shards stayed red, both diagnosed and neither caused by this branch:

- **Shard 5**: `conformance/test_opinfo.py`'s `div_floor_rounding`/
  `div_trunc_rounding` `_MATCHES_CPU` entries are stale (the cases now
  pass and the suite correctly says to delete them). This is fixed by
  open PR #409, which this branch predates -- no code change here;
  rebase onto it once it merges.
- **Shard 8**: `test_matches_cpu_nn_functional_conv2d_mojo_float32` --
  "Tensor-likes are not close! Mismatched elements: 1/256 (0.4%),
  greatest absolute difference 1.335e-05 at (0,3,3,3) (up to 1e-05
  allowed)", on a `padding='same'`, even-kernel, odd-dilation sample.
  `conformance/known_unsupported.py`'s `_ACCELERATOR_DELTAS["cpu"]`
  table -- "cpu" meaning a machine with no accelerator at all, i.e. this
  exact CI runner -- already declares `nn_functional_conv2d` unsupported
  on `test_matches_cpu` for bfloat16, float16 and int64: this repo's
  im2col+GEMM Mojo conv has known marginal numerical differences from
  PyTorch's independently-implemented CPU conv for those dtypes,
  predating this PR entirely. float32 simply was never in that list
  because `conv_ops.mojo` could not import on a no-accelerator host
  before c8dab40 -- there was nothing to measure. Confirmed this is the
  same class of pre-existing difference, not a regression: this machine
  cannot reproduce the CI failure directly (it has a real accelerator),
  but forcing `mojo:0` onto the CPU device the same way
  (`CUDA_VISIBLE_DEVICES=""`) and running the identical test passes 5/5
  on this branch and 3/3 on a clean upstream/main worktree -- the
  unmodified `_im2col_scalar`/`_bias_add_chan_scalar` CPU path behaves
  identically either side, consistent with `compare_kernel_asm.py`
  already having shown 0 device-code changes for `conv_ops`. float32
  conv never reaches the implicit-GEMM route (bf16-only) or the
  tensor-core/TF32 routes at default precision, so its accumulation is
  the same unmodified SIMT GEMM this branch never touched. Fixed (fa97f27)
  by extending the existing declaration to `"float32"`, the same
  mechanism already covering the other three dtypes of this operator.

## Open questions for the maintainer

These are policy decisions, not implementation questions, and this PR does
not resolve them on its own authority:

1. **Should this repo's benchmark suite reference `channels_last` instead
   of (or alongside) NCHW for `F.conv2d`?** This changes what "stock" means
   for every convolution baseline in `benchmarks/baselines.html`, not just
   this route's five nodes -- it is a suite-wide convention decision, and
   reasonable people can pick either (NCHW because it's what most callers
   actually hand `F.conv2d`, or channels_last because it's the layout
   cuDNN is fastest at and so the "fair" per-kernel comparison).
2. **Should this route's f32 gate follow `torch.backends.cudnn.allow_tf32`
   (cuDNN's actual default, `True`, so f32 conv gets TF32 unless the user
   opts out) instead of `torch.get_float32_matmul_precision()` (this
   repo's existing mm/bmm/conv-TF32 convention, default `"highest"`, TF32
   off)?** Matching cuDNN's own default would make the f32 numbers in this
   PR apples-to-apples without any kernel change; it would also be a new,
   conv-specific exception to the numerics policy every other TF32 gate in
   this repo shares.

### DIFF
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6415,7 +6415,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
-                    "contig": 1.311
+                    "contig": 1.074
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6408,14 +6408,49 @@ document.addEventListener("DOMContentLoaded", boot);
           "dtypes": {
             "bf16": {
               "shapes": {
+                "N16xC64x32x32_K64k3s1d2": {
+                  "layouts": {
+                    "contig": 6.791
+                  }
+                },
+                "N16xC96x28x28_K192k3s2": {
+                  "layouts": {
+                    "contig": 5.337
+                  }
+                },
+                "N1xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 1.806
+                  }
+                },
+                "N1xC64x56x56_K64k3s1": {
+                  "layouts": {
+                    "contig": 1.375
+                  }
+                },
+                "N32xC128x28x28_K128k3s2": {
+                  "layouts": {
+                    "contig": 12.566
+                  }
+                },
                 "N32xC256x14x14_K256k3s1": {
                   "layouts": {
                     "contig": 1.452
                   }
                 },
+                "N32xC256x14x14_K512k1s2": {
+                  "layouts": {
+                    "contig": 3.482
+                  }
+                },
                 "N32xC256x56x56_K64k1s1": {
                   "layouts": {
                     "contig": 0.783
+                  }
+                },
+                "N32xC512x7x7_K512k3s1": {
+                  "layouts": {
+                    "contig": 2.019
                   }
                 },
                 "N32xC64x56x56_K64k3s1": {
@@ -6428,6 +6463,11 @@ document.addEventListener("DOMContentLoaded", boot);
                     "contig": 1.102
                   }
                 },
+                "N8xC128x112x112_K128k3s1": {
+                  "layouts": {
+                    "contig": 0.981
+                  }
+                },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
                     "contig": 1.074
@@ -6437,14 +6477,49 @@ document.addEventListener("DOMContentLoaded", boot);
             },
             "f32": {
               "shapes": {
+                "N16xC64x32x32_K64k3s1d2": {
+                  "layouts": {
+                    "contig": 7.961
+                  }
+                },
+                "N16xC96x28x28_K192k3s2": {
+                  "layouts": {
+                    "contig": 7.757
+                  }
+                },
+                "N1xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 2.278
+                  }
+                },
+                "N1xC64x56x56_K64k3s1": {
+                  "layouts": {
+                    "contig": 4.026
+                  }
+                },
+                "N32xC128x28x28_K128k3s2": {
+                  "layouts": {
+                    "contig": 12.283
+                  }
+                },
                 "N32xC256x14x14_K256k3s1": {
                   "layouts": {
                     "contig": 19.806
                   }
                 },
+                "N32xC256x14x14_K512k1s2": {
+                  "layouts": {
+                    "contig": 4.427
+                  }
+                },
                 "N32xC256x56x56_K64k1s1": {
                   "layouts": {
                     "contig": 1.866
+                  }
+                },
+                "N32xC512x7x7_K512k3s1": {
+                  "layouts": {
+                    "contig": 18.407
                   }
                 },
                 "N32xC64x56x56_K64k3s1": {
@@ -6455,6 +6530,11 @@ document.addEventListener("DOMContentLoaded", boot);
                 "N7xC48x39x53_K80k3s1": {
                   "layouts": {
                     "contig": 2.87
+                  }
+                },
+                "N8xC128x112x112_K128k3s1": {
+                  "layouts": {
+                    "contig": 7.679
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6410,12 +6410,12 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
-                    "contig": 24.351
+                    "contig": 9.373
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
-                    "contig": 4.701
+                    "contig": 1.311
                   }
                 }
               }
@@ -6424,12 +6424,12 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
-                    "contig": 12.456
+                    "contig": 7.448
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
-                    "contig": 4.138
+                    "contig": 1.964
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6408,9 +6408,24 @@ document.addEventListener("DOMContentLoaded", boot);
           "dtypes": {
             "bf16": {
               "shapes": {
+                "N32xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 1.452
+                  }
+                },
+                "N32xC256x56x56_K64k1s1": {
+                  "layouts": {
+                    "contig": 0.783
+                  }
+                },
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
-                    "contig": 9.373
+                    "contig": 0.653
+                  }
+                },
+                "N7xC48x39x53_K80k3s1": {
+                  "layouts": {
+                    "contig": 1.102
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
@@ -6422,9 +6437,24 @@ document.addEventListener("DOMContentLoaded", boot);
             },
             "f32": {
               "shapes": {
+                "N32xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 19.806
+                  }
+                },
+                "N32xC256x56x56_K64k1s1": {
+                  "layouts": {
+                    "contig": 1.866
+                  }
+                },
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
                     "contig": 7.448
+                  }
+                },
+                "N7xC48x39x53_K80k3s1": {
+                  "layouts": {
+                    "contig": 2.87
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6390,9 +6390,24 @@ document.addEventListener("DOMContentLoaded", boot);
           "dtypes": {
             "bf16": {
               "shapes": {
+                "N32xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 1.452
+                  }
+                },
+                "N32xC256x56x56_K64k1s1": {
+                  "layouts": {
+                    "contig": 0.783
+                  }
+                },
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
-                    "contig": 9.373
+                    "contig": 0.653
+                  }
+                },
+                "N7xC48x39x53_K80k3s1": {
+                  "layouts": {
+                    "contig": 1.102
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
@@ -6404,9 +6419,24 @@ document.addEventListener("DOMContentLoaded", boot);
             },
             "f32": {
               "shapes": {
+                "N32xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 19.806
+                  }
+                },
+                "N32xC256x56x56_K64k1s1": {
+                  "layouts": {
+                    "contig": 1.866
+                  }
+                },
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
                     "contig": 7.448
+                  }
+                },
+                "N7xC48x39x53_K80k3s1": {
+                  "layouts": {
+                    "contig": 2.87
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6392,12 +6392,12 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
-                    "contig": 24.351
+                    "contig": 9.373
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
-                    "contig": 4.701
+                    "contig": 1.311
                   }
                 }
               }
@@ -6406,12 +6406,12 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "N32xC64x56x56_K64k3s1": {
                   "layouts": {
-                    "contig": 12.456
+                    "contig": 7.448
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
-                    "contig": 4.138
+                    "contig": 1.964
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6397,7 +6397,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
-                    "contig": 1.311
+                    "contig": 1.074
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -6390,14 +6390,49 @@ document.addEventListener("DOMContentLoaded", boot);
           "dtypes": {
             "bf16": {
               "shapes": {
+                "N16xC64x32x32_K64k3s1d2": {
+                  "layouts": {
+                    "contig": 6.791
+                  }
+                },
+                "N16xC96x28x28_K192k3s2": {
+                  "layouts": {
+                    "contig": 5.337
+                  }
+                },
+                "N1xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 1.806
+                  }
+                },
+                "N1xC64x56x56_K64k3s1": {
+                  "layouts": {
+                    "contig": 1.375
+                  }
+                },
+                "N32xC128x28x28_K128k3s2": {
+                  "layouts": {
+                    "contig": 12.566
+                  }
+                },
                 "N32xC256x14x14_K256k3s1": {
                   "layouts": {
                     "contig": 1.452
                   }
                 },
+                "N32xC256x14x14_K512k1s2": {
+                  "layouts": {
+                    "contig": 3.482
+                  }
+                },
                 "N32xC256x56x56_K64k1s1": {
                   "layouts": {
                     "contig": 0.783
+                  }
+                },
+                "N32xC512x7x7_K512k3s1": {
+                  "layouts": {
+                    "contig": 2.019
                   }
                 },
                 "N32xC64x56x56_K64k3s1": {
@@ -6410,6 +6445,11 @@ document.addEventListener("DOMContentLoaded", boot);
                     "contig": 1.102
                   }
                 },
+                "N8xC128x112x112_K128k3s1": {
+                  "layouts": {
+                    "contig": 0.981
+                  }
+                },
                 "N8xC3x224x224_K64k7s2": {
                   "layouts": {
                     "contig": 1.074
@@ -6419,14 +6459,49 @@ document.addEventListener("DOMContentLoaded", boot);
             },
             "f32": {
               "shapes": {
+                "N16xC64x32x32_K64k3s1d2": {
+                  "layouts": {
+                    "contig": 7.961
+                  }
+                },
+                "N16xC96x28x28_K192k3s2": {
+                  "layouts": {
+                    "contig": 7.757
+                  }
+                },
+                "N1xC256x14x14_K256k3s1": {
+                  "layouts": {
+                    "contig": 2.278
+                  }
+                },
+                "N1xC64x56x56_K64k3s1": {
+                  "layouts": {
+                    "contig": 4.026
+                  }
+                },
+                "N32xC128x28x28_K128k3s2": {
+                  "layouts": {
+                    "contig": 12.283
+                  }
+                },
                 "N32xC256x14x14_K256k3s1": {
                   "layouts": {
                     "contig": 19.806
                   }
                 },
+                "N32xC256x14x14_K512k1s2": {
+                  "layouts": {
+                    "contig": 4.427
+                  }
+                },
                 "N32xC256x56x56_K64k1s1": {
                   "layouts": {
                     "contig": 1.866
+                  }
+                },
+                "N32xC512x7x7_K512k3s1": {
+                  "layouts": {
+                    "contig": 18.407
                   }
                 },
                 "N32xC64x56x56_K64k3s1": {
@@ -6437,6 +6512,11 @@ document.addEventListener("DOMContentLoaded", boot);
                 "N7xC48x39x53_K80k3s1": {
                   "layouts": {
                     "contig": 2.87
+                  }
+                },
+                "N8xC128x112x112_K128k3s1": {
+                  "layouts": {
+                    "contig": 7.679
                   }
                 },
                 "N8xC3x224x224_K64k7s2": {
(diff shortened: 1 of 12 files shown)


gabrieldemarmiesse:
Fix conflicts

PR #391, opened 2026-08-16
FA4 forward: BHSD-native TMA descriptors eliminate the 3 gather-copies (H100)

gabrieldemarmiesse-ai-agent:
## Diagnosis

FA4 forward on H100 was 1.36-1.48x slower than stock PyTorch flash attention,
even though the FA4 kernel itself already matched or beat stock
kernel-vs-kernel on every measured shape. The gap was three
`data_movement_ops__run_gather_k` copies per call: `_fa4_native_bthd` in
`aten_fast.py` materializes Q/K/V from the public contiguous `(B, H, S, D)`
layout into `(B, S, H, D)` because the TMA descriptor path only accepted
`h_stride==64` layouts. Stock flash (FA2/FA3) consumes strided BHSD
directly — copies are 30-45% of the op, running at ~1.67 TB/s (already near
DRAM speed, not optimizable, only removable).

## The fix: a BHSD-native TMA descriptor variant

Added a `bhsd` comptime kernel param (`fa4_fwd_kernel.mojo`) and a matching
`bhsd_qkv` launch param (`fa4_fwd_launch.mojo`): Q/K/V/O TMA descriptors
address the **public** contiguous `(B, H, S, D)` tensor directly, viewed as
`(B*H, S, D)` planes — planes are the TMA outer dim and `S` is its own dim,
so the existing `BM=192` tail-tile zero-fill-on-load / clamp-on-store logic
handles every plane edge correctly with no BTHD materialization. This is
**FlashAttention-3's structure** (head stride is just another gmem stride);
none of FA3's scheduling machinery was needed since the existing FA4
pingpong/warp-specialized schedule already runs at stock parity. Both
params are additive with `False` defaults — every pre-existing instantiation
takes the exact same code path as before. Both bf16 and f16 are
instantiated (`flash_attention_fwd_{bf16,f16}_d64_causal_bhsd` in
`fa4_ops.mojo`), matching the existing bf16/f16 export-pair convention from
#386.

`fast_fa4_16bit_d64_causal_forward` in `aten_fast.py` now takes this path
whenever Q/K/V are all eligible (`_fa4_bhsd_layout`, see next section),
allocating the output directly in the same contiguous `(B, H, S, D)` layout
and returning it as-is — no transpose-back either. Ineligible views
(non-contiguous, or contiguous-but-misaligned) fall through to the existing
copy/strided paths unchanged.

## The alignment gate (review MUST-FIX)

Agent A2's review flagged that the Python-side eligibility gate must check
16-byte base-pointer alignment in addition to contiguity, since TMA
descriptor creation requires it and a sliced/offset bf16/f16 view can be
fully contiguous yet still violate it. `_fa4_bhsd_layout` in `aten_fast.py`
checks both, mirroring the existing `_fa4_strided_bthd_layout` precedent
for the strided BTHD ABI:

```python
return (
    tensor is not None
    and tensor._dtype in (DType.bfloat16, DType.float16)
    and tensor._itemsize == tensor._dtype.size_in_bytes
    and len(tensor._shape) == 4
    and tensor._is_contiguous
    and tensor._ptr % 16 == 0
)
```

Covered by three tests, one against a real device tensor:
- `tests/test_fa4_host_wiring.py::test_fa4_bhsd_layout_requires_contiguity_and_16_byte_alignment`
- `tests/test_fa4_host_wiring.py::test_fa4_offset_view_public_layout_copies_and_uses_contiguous_fallback` (mocked, repurposed from the pre-existing "unsupported public layout" test, which is what a plain 16-byte-aligned contiguous tensor now *is* eligible for)
- `tests/test_eager_kernels.py::test_fa4_bhsd_gate_rejects_misaligned_offset_view` (real H100 tensor: `storage[1:].view(...)`, spies confirm the dense bridge runs and the bhsd bridge never does, output still matches the CPU reference)

## Backward: no interaction

No backward kernel/launch changes. Verified by reading
`fast_aten__scaled_dot_product_flash_attention_backward`: PyTorch's
generated autograd formula for `_scaled_dot_product_flash_attention` saves
the **public** `query/key/value/output/logsumexp` tensors (not whatever
forward's `q_native/k_native/v_native` return values were), and backward
always re-derives its own BTHD natives from those public tensors via
`_fa4_native_bthd`. I also simplified backward's rare
SavedVariable-recompute fallback branch (triggered when PyTorch's autograd
returns `output`/`logsumexp` as bare tensors without the payload) to always
re-derive natives independently rather than trusting forward's returned
tuple — the new BHSD fast path returns the public q/k/v *unchanged* (that's
the entire point, no BTHD copy), which is the wrong shape for the
backward-only BTHD ABI, so this path can no longer assume forward already
built usable natives.

One real, documented nuance: backward's `_fa4_native_bthd(out)` used to get
a free transposed view of the physical BTHD `out` (double-transpose
cancels); now that `out` from the BHSD fast path is already public-shaped
and contiguous, backward pays one `O` copy (~15us on P0) where forward
used to pay 3 copies (~45us) — net win in training too. `bf16` backward's
recorded baseline (0.716) didn't move outside the 8% dead band in this
PR's measurement, confirming zero backward regression.

## Per-node before/after (H100 PCIe, pinned 1395MHz, device time, B8H12S1024D64)

| aten op (dtype) | before (recorded baseline) | after (this PR) |
|---|---|---|
| `_scaled_dot_product_flash_attention` (bf16) | 1.365 | **0.888** |
| `_scaled_dot_product_flash_attention` (f16) | *(not previously recorded)* | **0.883** |
| `_scaled_dot_product_efficient_attention` (bf16) | 0.724 | **0.477** |
| `_scaled_dot_product_efficient_attention` (f16) | 3.461 | **0.478** |
| `scaled_dot_product_attention` (composite, bf16) | 2.255 | **1.468** |
| `scaled_dot_product_attention` (composite, f16) | 11.073 | **1.464** |
| `_scaled_dot_product_flash_attention_backward` (bf16) | 0.716 | not flagged as changed (within the +/-8% dead band; passed both runs, no update written) |
| `_scaled_dot_product_flash_attention_backward` (f16) | *(not previously recorded)* | **0.764** |

`_scaled_dot_product_flash_attention` and `_scaled_dot_product_efficient_attention`
both clear the 10% bar. All numbers recorded via
`uv run pytest benchmarks/test_attention.py --update-baselines`;
`benchmarks/baselines.html` is included in this PR, touching only these 7
entries (`git diff --stat`: 23 insertions, 5 deletions).

Cross-checked against the harness's pure-Mojo numbers in
`/scratch/fa4-fwd-harness/NOTES.md` (P0 shape, streamed device time,
checks on): base(BTHD kernel) 98.0us, candidate(BHSD) 96.1us, stock
109.8us, cand/stock 0.875 — this PR's end-to-end op-level 0.888 (bf16)
matches that kernel-level number closely once ~9us of Python/launch
overhead on a 97us op is folded back in.

## Phase-2 gap (documented, not fixed here)

The composite `scaled_dot_product_attention` node improves a lot (bf16:
2.255 -> 1.468; f16: 11.073 -> 1.464) but stays above the 1.10 bar: stock's
leg there is cuDNN's fused attention kernel, not `flash_fwd_kernel`, and
closing that gap is unrelated to FA4's own efficiency (FA4 itself is
already within 10% of stock's `_scaled_dot_product_flash_attention` path).
Out of scope for this PR.

## Blast radius (`scripts/compare_kernel_asm.py`)

- **sm_90a**: `fa4_ops` module goes from 10 to 12 kernels. All 6 backward
  kernels (preprocess/main/convert x2 dtypes) are byte-identical (exact
  mangled-hash match, not just masked-hash). All 4 pre-existing forward
  kernels (dense/strided x2 dtypes) each differ by exactly one cosmetic
  token — an LLVM module-scope `extern_ptr_syml_N` symbol ordinal that
  shifted by +2 because the new bhsd launch call site sits earlier in
  `fa4_fwd_launch.mojo` — confirmed by masking the mangling hash and diffing
  each pair: 12 lines differ, all on that one token, zero instruction
  changes (verified for all 4 pairs, matched by exact line count and
  content). The remaining 2 kernels are the new bf16/f16 bhsd
  instantiations. **Zero functional movement on any pre-existing sm_90a
  kernel.**
- **gfx942**: `fa4_ops` fails to build identically on both pristine
  `upstream/main` and this branch (`error: failed to run the pass manager
  for offload functions`) — FA4 uses Hopper TMA/wgmma intrinsics
  (`max.gpu.host.nvidia.tma`) and has never targeted AMD. **Zero gfx942
  footprint, before and after, identical failure signature.**

## Tests

- `tests/test_fa4_host_wiring.py`: 15 tests, including the new
  `_fa4_bhsd_layout` unit test, the repurposed offset-view fallback test,
  and a new positive-path test proving the bhsd bridge is reached (and the
  old materialize/copy path is *not*) for aligned contiguous public Q/K/V.
- `tests/test_eager_kernels.py`: ported agent A's 7 pure-Mojo edge shapes
  (every `S % BM=192` tail-residue class reachable under the
  `seqlen % 128 == 0` gate) into a real forward+backward SDPA test against
  a CPU fp32 reference, `x2` dtypes = 14 cases, plus the real-GPU
  misaligned-offset-view fallback test.
- Full FA4/SDPA-relevant selection across
  `test_eager_kernels.py`/`test_fa4_host_wiring.py`/`test_aten_functions.py`/
  `test_high_level_ops.py`/`test_mojo_autocast.py`/`test_compiler.py`/
  `test_call_queue.py`/`test_tensor_spec_oom.py`: **113 passed, 8 skipped
  (unrelated pre-existing skips), 0 failed.**
- `benchmarks/test_coverage.py`: 2 passed (op-registration and
  baseline-addressability reconciliation, both GPU-free).

## Credit

The bhsd descriptor structure follows FlashAttention-3's approach of
treating head stride as just another gmem stride rather than a distinct TMA
dimension, applied here to FA4's existing warp-specialized pingpong
schedule with no scheduling changes.

---

### Gap

`B1H16S4096D128` (head_dim=128) was declining FA4's `head_dim == 64` gate
for both dtypes and falling all the way to the bmm+softmax decomposition:
composite/efficient SDPA measured **~10.6x / ~3.1x** slower than stock
there — this is exactly [A2 in `docs/optimization_backlog.md`](docs/optimization_backlog.md).

### What changed

The FA4 kernel family turned out to already be comptime-parametric over
`head_dim` in the places that matter: `fa4_fwd_common.mojo` /
`fa4_bwd_common.mojo` already carried BM, MMA-warpgroup count, register
budgets and the backward causal tile_m as `head_dim`-parametric constants
mirroring Tri Dao FA4's own sm90 configs (BM=128 not 192, 2 MMA
warpgroups not 3, bwd causal tile_m=64 not 128) — this was written when
FA4 was first vendored but never exported past `head_dim == 64` while this
PR's own scope stayed narrow. The gap was pure wiring, not a missing
kernel:

- **`fa4_ops.mojo`**: exported the same three forward entry points (dense,
  `_strided_qkv` zero-copy BTHD, `_bhsd` native) and two backward entry
  points (dense, `_strided_qkv`) for `d128`, mirroring the `d64` family
  1:1 for both bf16 and f16 — 10 new `flash_attention_{fwd,bwd}_*_d128_*`
  Python-callable symbols. `_check_strided_qkv_layout` now takes `head_dim`
  as a parameter instead of hardcoding 64 in its `h_stride`/`s_stride`
  checks.
- **`fa4_fwd_launch.mojo`**: widened the `strided_qkv`/`bhsd_qkv` comptime
  asserts from `head_dim == 64` to `(head_dim == 64 or head_dim == 128)`.
  The strided-Q TMA descriptor branch existed only inside the `QO_RANK ==
  4` path (d64's BM=192 rank-4 tail-tile fix); d128 uses `QO_RANK == 3`
  (BM=128 always divides evenly under the `seqlen % 128 == 0` gate, so it
  never needed that fix) and was building a plain, non-strided Q
  descriptor unconditionally — added the missing `comptime if
  strided_qkv:` branch there too, mirroring the pattern K/V already used.
- **`fa4_bwd_launch.mojo`**: widened the matching backward assert; Q/K/V
  descriptor construction in `launch_bwd_main` was already
  `head_dim`-generic (no `QO_RANK`-style branch to extend).
- **`aten_fast.py`**: `_fa4_16bit_d64_causal_inputs` and
  `_fa4_strided_bthd_layout` now accept `head_dim in (64, 128)` (the
  latter's `h_stride`/`s_stride` checks scale with `head_dim` instead of
  a hardcoded 64). `_fa4_bhsd_layout` needed no change — it was already
  `head_dim`-generic. A new `_fa4_symbol()` helper derives the
  `bf16/f16 x d64/d128 x dense/strided/bhsd` Mojo symbol name from
  `head_dim` at each of the 5 call sites that used to hardcode a
  `d64`-suffixed ternary on dtype alone.

Backward is included (not scoped out): `fa4_bwd_common.mojo`'s own
docstring already documented the d128 causal tile_m (64) alongside d64's
(128), so the backward kernel bodies were exported and tested exactly like
forward — no backward-specific tile work was needed.

### Tests

Parametrized the existing FA4 GPU test suite over `head_dim in {64, 128}
x {bf16, f16}` (40 cases total, `tests/test_eager_kernels.py`):

- `test_fa4_causal_gapped_qkv_forward_backward` — nanoGPT-style gapped
  fused-QKV views, strided zero-copy forward+backward.
- `test_fa4_direct_flash_aten_returns_real_logsumexp` — low-level flash op
  LSE/gradient correctness.
- `test_fa4_bhsd_native_forward_backward_matches_reference` — the 7
  BM-tail-residue edge shapes (single m-block, exact multiples, odd
  plane counts, LPT-swizzle-triggering sizes) x head_dim x dtype. At
  d128, BM==128 always divides the `seqlen % 128 == 0` gate evenly (no
  partial m-tile is reachable there, unlike d64's BM=192), so this
  doubles as a multi-plane/LPT-scheduling stress set at the second tile
  config rather than a literal tail-residue test.
- `test_fa4_bhsd_gate_rejects_misaligned_offset_view` — the 16-byte
  alignment fallback (dense/copy path), at both head_dims.

All 40 pass. The broader FA4/SDPA-relevant regression sweep across
`test_eager_kernels.py`, `test_fa4_host_wiring.py`, `test_aten_functions.py`,
`test_high_level_ops.py`, `test_mojo_autocast.py`: **141 passed, 8 skipped
(pre-existing, unrelated), 0 failed.** `test_fa4_host_wiring.py`'s 16
host-only tests (no GPU) also pass unchanged — `_fa4_symbol`'s type hint
had to move from the (more precise but too strict) `ModuleType` to
`object` so it still accepts the `SimpleNamespace` doubles those tests
mock `load_fa4_ops()` with.

### Before/after (H100 PCIe, pinned clock, device time, `B1H16S4096D128`)

| aten op (dtype) | before | after |
|---|---|---|
| `_scaled_dot_product_flash_attention` (bf16) | *(skip — declined FA4's gate)* | **0.413** (new entry) |
| `_scaled_dot_product_flash_attention` (f16) | *(skip)* | **0.417** (new entry) |
| `_scaled_dot_product_flash_attention_backward` (bf16) | *(skip)* | **0.592** (new entry) |
| `_scaled_dot_product_flash_attention_backward` (f16) | *(skip)* | **0.593** (new entry) |
| `_scaled_dot_product_efficient_attention` (bf16) | 3.147 | **0.249** |
| `_scaled_dot_product_efficient_attention` (f16) | 3.137 | **0.251** |
| `scaled_dot_product_attention` (composite, bf16) | 10.750 | **0.832** |
| `scaled_dot_product_attention` (composite, f16) | 10.649 | **0.835** |

Every node clears the 1.10 bar comfortably — most of them are now
*faster* than stock PyTorch at this shape, not just within 10% of it
(e.g. flash forward: 180us ours vs 436us stock). `B8H12S1024D64` (the
canary) was re-verified read-only (no `--update-baselines`): all 10 nodes
pass unchanged, confirmed by diffing `baselines.html` — the touched JSON
keys are exclusively under `B1H16S4096D128` (28 lines changed total: 4
value updates + 4 new entries, times two ops each for fwd/bwd... see the
commit for the exact diff). Recorded via
`uv run pytest benchmarks/test_attention.py -k B1H16S4096D128
--update-baselines`; `benchmarks/test_coverage.py` (op-registration +
baseline-addressability reconciliation) passes.

### Blast radius (`scripts/compare_kernel_asm.py`)

- **sm_90a**: `fa4_ops` goes from 12 to 22 kernels (+10, 0 removed): 4 new
  forward bodies + 6 new backward bodies (preprocess/main/convert x2
  dtypes), all exclusively reachable through the new d128 entry points.
  The 10 pre-existing d64 kernels each differ by exactly one cosmetic
  token — the same `extern_ptr_syml_N` LLVM module-scope symbol-ordinal
  shift documented in this PR's own earlier blast-radius section above,
  recurring because the new d128 call sites sit earlier in
  `fa4_ops.mojo`/`fa4_fwd_launch.mojo`. Full `--show-diff` confirms zero
  instruction-level changes on every pre-existing kernel. **Zero
  functional movement on any pre-existing sm_90a kernel.**
- **gfx942**: fails identically before and after
  (`error: failed to run the pass manager for offload functions` — same
  Hopper-only TMA/wgmma dependency as before). **Zero gfx942 footprint,
  identical failure signature.**

---

## Forward schedule (phase 2b)

Follow-up to the head_dim=128 section above: closes the gap between FA4
fwd (post-bhsd, `a4e8462`) and **cuDNN attention**
(`sdpa_kernel(CUDNN_ATTENTION)`), the backend stock PyTorch actually picks
on H100 — 1.6x faster than the FLASH backend this PR's earlier sections
benchmarked against. Integrates agent A's phase-2/2b forward-schedule
harness (`/scratch/fa4-fwd-harness`, `NOTES.md`).

### Root cause and fix

`BM=192 > BN=128` makes the causal diagonal band span two 128-column
tiles per 192 rows: at `S=1024` the kernel computed **1.500x** the causal
triangle vs **1.125x** at `BM=64` (cuDNN's own tile is 64x128). d64 moves
tile `192x128` (3 MMA warpgroups, 512 threads, 1 CTA/SM) to `64x128` (1
MMA warpgroup, 256 threads, **2 CTAs/SM** via `nvvm.minctasm=2` — the
second CTA replaces the third warpgroup's latency hiding). On top of the
geometry, an instruction diet on the causal mask and row reductions (see
the kernel-level commit for the full list): 32-bit mask thresholds,
exact per-warpgroup diagonal predicate, column-group skip, tree-shaped
rowmax/rowsum, and dropping the now-self-ring pingpong barrier at
`NWG==1`. `head_dim=128` keeps `a4e8462`'s tile/regs/codegen throughout
(every new axis is comptime-gated off there).

### Per-shape harness table (streamed device time, us, pinned 1395MHz, single-shape processes)

| shape | v1 = a4e8462 | v2 (phase 2, geometry only) | v2b (phase 2b, + diet) | cuDNN | v2b/cuDNN | bar 1.10 |
|-------|--------------|------------------------------|-------------------------|-------|-----------|----------|
| P0 B8 H12 S1024  | 94.5-97.5   | 73.2-75.1   | 70.0-71.1   | 66.0  | 1.06-1.08 | PASS |
| P1 B32 H12 S1024 | 348.7-350.3 | 284.1-284.6 | 261.9-262.9 | 227.1 | **1.15-1.16** | **FAIL (known residual, see below)** |
| P2 B1 H16 S4096  | 127.7-130.1 | 131.3-133.7 | 127.6-129.3 | 151.4 | 0.84-0.85 | PASS |
| P3 B16 H16 S2048 | 613-622     | 587-596     | 555.6-569.3 | 508.0 | 1.09-1.12 | PASS (at the bar) |
| P4 B3 H5 S1152   | 24.5-25.1   | 17.6-17.9   | 16.7-16.9   | 21.6  | 0.77-0.78 | PASS |
| P5 B1 H8 S512    | 15.0-15.5   | 9.25-9.69   | 8.47-8.75   | 9.8   | 0.86-0.89 | PASS |

vs the branch starting point (v1): P0 -27%, P1 -25%, P3 -9%, P4 -33%,
P5 -44%. d128 guard (`B8H12S1024D128` / `B32H12S1024D128`, ncu A/A
controlled): many-wave shape 9% faster (the phase-2 wave-dispatched
swizzle applying to d128 for the first time), low-wave shape neutral
(apparent 3% is inside the harness's own 3.8% A/A leg-asymmetry bound);
bitwise-identical outputs both shapes.

### Benchmark-node before/after (`benchmarks/test_attention.py`, H100 PCIe, pinned 1395MHz, device time)

**D64 (B8H12S1024D64) — moves, mirrors the harness's v1->v2b deltas:**

| aten op (dtype) | before (branch head) | after (this change) |
|---|---|---|
| `scaled_dot_product_attention` (bf16) | 1.468 | **1.066** |
| `scaled_dot_product_attention` (f16) | 1.464 | **1.061** |
| `_scaled_dot_product_flash_attention` (bf16) | 0.888 | **0.643** |
| `_scaled_dot_product_flash_attention` (f16) | 0.883 | **0.635** |
| `_scaled_dot_product_efficient_attention` (bf16) | 0.477 | **0.354** |
| `_scaled_dot_product_efficient_attention` (f16) | 0.478 | **0.355** |
| `_scaled_dot_product_flash_attention_backward` (bf16) | 0.716 | unchanged (within +/-8% dead band; own constants, untouched file) |

The composite `scaled_dot_product_attention` node (stock leg = cuDNN)
crosses the 1.10 bar for the first time on this shape: 1.066/1.061,
matching the harness's P0 v2b/cuDNN of 1.06-1.08 once Python/launch
overhead on a ~70us op is folded back in.

**D128 (B1H16S4096D128) — stable, as expected (this diff's new axes are
comptime-gated off at d128; only `.minnctapersm 1` is emitted there, see
blast radius):**

| aten op (dtype) | before | after |
|---|---|---|
| `_scaled_dot_product_flash_attention` (bf16/f16) | 0.413 / 0.417 | unchanged |
| `_scaled_dot_product_efficient_attention` (bf16/f16) | 0.249 / 0.251 | unchanged |
| `scaled_dot_product_attention` (bf16/f16) | 0.832 / 0.835 | unchanged |
| `_scaled_dot_product_flash_attention_backward` (bf16/f16) | 0.592 / 0.593 | unchanged |

All 20 `test_attention.py` cases pass; `benchmarks/baselines.html` is
included, touching exactly the 6 D64 entries above (`git diff --stat`:
6 value updates, nothing else).

### Known residual: P1 (B32 H12 S1024) at 1.15-1.16

Per-unit accounting (ncu, 27 waves, 2 CTAs/SM): after the diet, the
kernel executes ~610 warp-instructions per 64x128 unit against cuDNN's
~621 (instruction parity) but issues 265k inst/us against their 304k
(-13%) — the gap is issue efficiency / latency hiding with only 2
consumer warps per SMSP and `wait` (fixed-latency dependency) as the top
stall, not either compute pipe (tensor 36%, XU/exp2 39% of peak). P1's
100 MB K/V working set is also the one L2-nonresident shape here (L2 hit
71.5%, vs P0's 25 MB fitting easily) — cuDNN's persistent grid with a
dynamic atomic-counter scheduler is the plausible reason they gain more
per plane going P0->P1 (15%) than this kernel does (6.6%). **Phase 2c resolved this residual — see the
"Forward schedule (phase 2c: self-loading warpgroup)" section below; P1
now measures 1.02-1.05 in the harness and the composite benchmark node
sits at 0.912.**

### Dead ends (measured, so the next agent does not re-explore them)

- **BM=192 + per-warpgroup causal trip counts** (the natural first idea:
  keep 3 warpgroups, let each stop at its own causal boundary) —
  work-per-CTA fell 23% but wall time only -8%: a BM=192 CTA is bound by
  its K/V tile *stream*, not by warpgroup math, and the savings land on
  the tail m-block, which still streams every K/V tile for the CTA's
  other (still-active) warpgroups.
- **BM=128 / 2 warpgroups / 384 threads** — worse per-unit than 3
  warpgroups on long-seq shapes: each warpgroup's ~850-cycle softmax
  overruns its 512-cycle tensor slot at NWG=2, where NWG=3 had 2x512
  cycles of slack to hide it in.
- **8 K/V ring stages vs 6** — no measurable help at BM=128; reverted
  before BM=64 was tried (6 is what fits 2 CTAs/SM's smem budget there
  anyway).
- **L2-swizzle group size fitted to `2*sm_count/num_m_tiles`** — loses
  both ways vs either extreme (worse than group=1 AND worse than the
  production 50MB-budget formula); the shipped rule is a coarse
  wave-count dispatch (>=12 waves -> group 1, else the existing formula)
  instead.
- **Static persistent grid scheduler** (3 variants: single-buffered Q,
  double-buffered Q at 4 stages) — delivered its memory promise (DRAM
  reads 246->162 MB, L2 hit 50->75% at P1) and still lost ~3% wall time
  everywhere: round-robin static assignment forfeits the hardware
  scheduler's dynamic CTA placement (no work-stealing at the tail), and
  the loop-carried scheduler state spills registers the one-shot kernel
  never did. A *dynamic* (atomic-counter) persistent scheduler a la FA3
  might flip the sign but was not built.
- **Rowsum FADD removal via an MMA identity trick** and **packed-half
  softmax** — both measured negative (tensor is the tighter pipe already;
  packing added conversion overhead without freeing a bottlenecked unit).

### Blast radius (`scripts/compare_kernel_asm.py`)

- **sm_90a** vs `a4e8462`: `fa4_ops` still 22 kernels. The 6 d64 forward
  bodies (dense/strided_qkv/bhsd x bf16/f16) get new mangled names
  entirely (BM/thread-count/register-budget template arguments changed,
  not just their content) — 6 old symbols gone, 6 new symbols appear, as
  expected; this *is* the change. The 4 d128 forward bodies each gain
  exactly one line, `.minnctapersm 1` (from the new, now-unconditional
  `nvvm.minctasm` metadata — `2 if NWG==1 else 1` evaluates to `1` at
  d128's NWG=2, which is a no-op vs the previous absence of that
  attribute), zero other bytes differ; the harness's own A/A control
  bounds this line's effect at <=1% (noise), and the benchmark table
  above confirms d128 nodes did not move. All 12 backward kernels
  (`fa4_bwd_kernel.mojo`, its own constants, file untouched by this
  change) are byte-identical.
- **gfx942**: fails to build identically before and after (`error: failed
  to run the pass manager for offload functions` — the same Hopper-only
  TMA/wgmma dependency as every other FA4 change on this PR). Zero gfx942
  footprint.

### Tests

- Full FA4/SDPA selection (`tests/test_eager_kernels.py -k "fa4 or flash
  or sdpa"`, both dtypes, both head_dims): **93 passed, 4 skipped**
  (pre-existing, platform-gated).
- `tests/test_fa4_host_wiring.py`: **16 passed** (host-only, no GPU).
- New: `test_fa4_bhsd_d64_direct_partial_tail_block` (`S in {65, 127,
  193}`, both dtypes, 6 cases) — the permanent-suite home for the
  harness's genuine BM=64 partial-tail shapes. The production
  `seqlen % 128 == 0` eligibility gate would route every one of these
  away from FA4 (128 is now a multiple of BM=64, so the pre-existing
  `_FA4_BHSD_TAIL_SHAPES` set no longer reaches a partial last m-block at
  all under that gate — comments there updated to say so), so this test
  calls the compiled bhsd bridge directly, the same way
  `fast_fa4_16bit_d64_causal_forward` does internally, past the gate.
- `benchmarks/test_coverage.py`: 2 passed (GPU-free).
- `uvx pre-commit run` on the touched files: clean.

## Forward schedule (phase 2c: self-loading warpgroup)

The final forward commit replaces, for dense-causal BHSD d64 at >=2 waves,
the producer/consumer pair with a **single self-loading 128-thread
warpgroup** (each warpgroup refills the ring slot it just consumed, so the
`empty[]` barriers are deleted outright — `wgmma.wait_group` orders the
refill, the same release pattern CUTLASS `sm90_mma_tma_gmma_ss.hpp`, FA3's
mainloop, and MAX's own Hopper GEMM rely on) at **3 CTAs/SM** (154 regs,
4-stage 72 KiB ring). Below 2 waves the launcher keeps the phase-2b kernel
(wave count derived from the runtime SM count), which also preserves the
small-shape wins.

Phase 2c also **falsified the obvious idea first**: 2-CTA clusters with TMA
multicast were fully implemented, delivered the promised -33% L2 read
sectors, and ran +9% *slower* — the kernel is latency-bound, not L2-bound
(nothing was saturated: tensor 36%, XU 39%, L2 42%, issue 44% of peak). The
win is occupancy: warps active +21%, issue rate +14%, time -13.6% (ncu).

Harness numbers (streamed device time vs cuDNN attention, single-shape
processes, pinned 1395 MHz):

| shape | phase 2b | selfload | vs cuDNN |
|---|---|---|---|
| B8 H12 S1024 | 70.9-71.2 µs | 60.8-61.5 µs | **0.91** |
| B32 H12 S1024 | 264.9-269.9 | 229.8-236.0 | **1.02-1.05** |
| B1 H16 S4096 | 129.6-130.7 | 120.8-121.6 | **0.80** |
| B16 H16 S2048 | 570-575 | 520.6-525.7 | **1.03-1.04** |
| B3 H5 S1152 | 16.6-16.9 | 18.2-18.6 (v2b kernel via wave gate) | 0.78-0.87 |
| B1 H8 S512 | 8.5-8.9 | 9.0-9.3 (v2b kernel via wave gate) | 0.86-0.93 |

Recorded benchmark nodes after this commit (bf16/f16): composite
`scaled_dot_product_attention` D64 **0.912/0.907**, flash direct
**0.558/0.552**, efficient **0.308/0.310**; D128 and all backward nodes
unchanged within the dead band.

Correctness carried the heaviest review of the series: the barrier-deletion
argument was adversarially reviewed (verdict: sound, with the
prefetch-distance invariant now documented in the kernel next to the
CUTLASS/FA3/MAX precedent citations), a PTX-ordering regression test guards
against toolchain drift (Mojo's `wait_group` carries no memory clobber), a
2048-launch soak at three clock regimes ran bit-clean, and a CI-sized soak
plus selfload-reaching partial-tail tests are in the permanent suite.

## Seqlen gate relaxation (review thread follow-up)

The review thread asked to relax the `seqlen % 128 == 0` eligibility gate
(`_fa4_16bit_d64_causal_inputs`), a leftover from the pre-phase-2b `BM=192`
era. Landed for the bhsd forward route only, per-route policy:

| route | grad needed | seqlen requirement |
|---|---|---|
| BHSD-native forward (`fast_fa4_16bit_d64_causal_forward`, d64 and d128) | no | any `seqlen >= 1` |
| Legacy strided-BTHD / dense-copy forward | either | `seqlen % 128 == 0` (unproven tail machinery) |
| Any route, backward | yes | `seqlen % 128 == 0` (unproven tail machinery) |

The BHSD-native path's zero-fill/clamp tail machinery was already proven
correct at both d64 and d128 by the existing direct-bridge tests; this
change only widens what the *public* eligibility gate lets reach it.
`_fa4_16bit_d64_causal_inputs` grows an `allow_any_seqlen` kwarg (default
`False`, so every pre-existing call site — including both backward-adjacent
gates — keeps its exact old contract byte-for-byte); only
`fast_fa4_16bit_d64_causal_forward` passes `True`, and it still declines a
non-multiple-of-128 seqlen when the BHSD layout doesn't apply, so the
legacy routes stay exactly as conservative as before.

Backward was investigated and left untouched on purpose. The bwd kernels
address Q/K/V/dQ/dK/dV through one TMA descriptor per tensor spanning the
whole flattened `(B*S, H, D)` range (`fa4_bwd_launch.mojo`'s
`create_split_tma(ctx, ptr, rows=batch*seqlen, nheads)`), not one descriptor
per batch. A batch's partial last KV tile computes an all-zero
gradient contribution for its padding rows (causal masking already excludes
them; verified by reading `fa4_bwd_kernel.mojo`), but the TMA *store* of
that all-zero tile is not clipped at the batch boundary the way it is at
the very end of the tensor — for `batch > 1` it lands on the valid opening
rows of the *next* batch's `dK`/`dV`, racing that batch's own correct CTA
store. That is a real silent-wrong-gradient risk, not a hypothetical one,
so backward keeps requiring `seqlen % 128 == 0` everywhere, and a
training-shaped call (grad enabled, `requires_grad` Q/K/V) at an odd seqlen
falls through — via the autograd wrapper's unchanged `needs_backward` gate
— to the existing math-decomposition `_ScaledDotProductAttentionAutograd`
Function instead of the native flash forward/backward pair. Covered by
`test_fa4_sdpa_odd_seqlen_requires_grad_uses_decomposition`, which spies on
every FA4 forward/backward symbol and asserts none of them ran, and checks
the decomposition's gradients against a CPU fp64→fp32 reference.

New aten-level tests (`tests/test_eager_kernels.py`, through
`torch.nn.functional.scaled_dot_product_attention` and the direct
`aten::_scaled_dot_product_flash_attention` op — not the compiled bridge):
odd seqlens `{65, 127, 193, 1023, 4095} x {bf16, f16} x {d64, d128}` against
a CPU fp32 reference, each asserting (via a monkeypatch spy on the bhsd
symbol) that FA4 actually ran rather than merely producing matching numbers
through some other path.

**A/B spot-check** (not a recorded `benchmarks/` node — odd seqlens aren't
in its shape tables): kineto device time, B8H12 D64 causal, H100 PCIe
pinned at 1395 MHz, via `bench_lib.measure` (the same ABBA interleaved
methodology the recorded baselines use):

| dtype | mojo S1023 / mojo S1024 (tail-cliff check) | ours/stock @ S1023 | ours/stock @ S1024 |
|---|---|---|---|
| bf16 | **1.002** ± 0.4% (59.9us / 59.7us) | 0.891 ± 0.1% | 0.924 ± 0.4% |
| f16 | **0.999** ± 0.3% (60.3us / 60.4us) | 0.898 ± 0.0% | 0.922 ± 0.1% |

The odd seqlen lands within noise of the even one — one partial tile out of
16 costs nothing measurable, not a cliff — and both stay ~8-11% faster than
stock PyTorch, consistent with the recorded S1024 baseline.

**Cross-compilation**: `scripts/compare_kernel_asm.py --before <pre-change
worktree> --after . --accelerator sm_90a --kernel-dir
torch_mojo_backend/eager_flash_attention --modules fa4_ops` reports **0 of
24 kernels differ** — expected, since this change touches no `.mojo` source
at all (Python gate only).

`uvx pre-commit run --all-files`: clean.

### DIFF
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -5075,7 +5075,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.724
+                    "contig": 0.477
                   }
                 }
               }
@@ -5089,7 +5089,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 3.461
+                    "contig": 0.478
                   }
                 }
               }
@@ -5102,7 +5102,16 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 1.365
+                    "contig": 0.888
+                  }
+                }
+              }
+            },
+            "f16": {
+              "shapes": {
+                "B8H12S1024D64": {
+                  "layouts": {
+                    "contig": 0.883
                   }
                 }
               }
@@ -5116,6 +5125,15 @@ document.addEventListener("DOMContentLoaded", boot);
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.716
+                  }
+                }
+              }
+            },
+            "f16": {
+              "shapes": {
+                "B8H12S1024D64": {
+                  "layouts": {
+                    "contig": 0.764
                   }
                 }
               }
@@ -8584,7 +8602,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 2.255
+                    "contig": 1.468
                   }
                 }
               }
@@ -8598,7 +8616,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 11.073
+                    "contig": 1.464
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -5075,7 +5075,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.354
+                    "contig": 0.308
                   }
                 }
               }
@@ -5089,7 +5089,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.355
+                    "contig": 0.31
                   }
                 }
               }
@@ -5107,7 +5107,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.643
+                    "contig": 0.558
                   }
                 }
               }
@@ -5121,7 +5121,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.635
+                    "contig": 0.553
                   }
                 }
               }
@@ -8622,7 +8622,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 1.066
+                    "contig": 0.912
                   }
                 }
               }
@@ -8636,7 +8636,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 1.061
+                    "contig": 0.915
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -5075,7 +5075,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.477
+                    "contig": 0.354
                   }
                 }
               }
@@ -5089,7 +5089,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.478
+                    "contig": 0.355
                   }
                 }
               }
@@ -5107,7 +5107,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.888
+                    "contig": 0.643
                   }
                 }
               }
@@ -5121,7 +5121,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 0.883
+                    "contig": 0.635
                   }
                 }
               }
@@ -8622,7 +8622,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 1.468
+                    "contig": 1.066
                   }
                 }
               }
@@ -8636,7 +8636,7 @@ document.addEventListener("DOMContentLoaded", boot);
                 },
                 "B8H12S1024D64": {
                   "layouts": {
-                    "contig": 1.464
+                    "contig": 1.061
                   }
                 }
               }
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -5070,7 +5070,7 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "B1H16S4096D128": {
                   "layouts": {
-                    "contig": 3.147
+                    "contig": 0.249
                   }
                 },
                 "B8H12S1024D64": {
@@ -5084,7 +5084,7 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "B1H16S4096D128": {
                   "layouts": {
-                    "contig": 3.137
+                    "contig": 0.251
                   }
                 },
                 "B8H12S1024D64": {
@@ -5100,6 +5100,11 @@ document.addEventListener("DOMContentLoaded", boot);
           "dtypes": {
             "bf16": {
               "shapes": {
+                "B1H16S4096D128": {
+                  "layouts": {
+                    "contig": 0.413
+                  }
+                },
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.888
@@ -5109,6 +5114,11 @@ document.addEventListener("DOMContentLoaded", boot);
             },
             "f16": {
               "shapes": {
+                "B1H16S4096D128": {
+                  "layouts": {
+                    "contig": 0.417
+                  }
+                },
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.883
@@ -5122,6 +5132,11 @@ document.addEventListener("DOMContentLoaded", boot);
           "dtypes": {
             "bf16": {
               "shapes": {
+                "B1H16S4096D128": {
+                  "layouts": {
+                    "contig": 0.592
+                  }
+                },
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.716
@@ -5131,6 +5146,11 @@ document.addEventListener("DOMContentLoaded", boot);
             },
             "f16": {
               "shapes": {
+                "B1H16S4096D128": {
+                  "layouts": {
+                    "contig": 0.593
+                  }
+                },
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.764
@@ -8597,7 +8617,7 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "B1H16S4096D128": {
                   "layouts": {
-                    "contig": 10.75
+                    "contig": 0.832
                   }
                 },
                 "B8H12S1024D64": {
@@ -8611,7 +8631,7 @@ document.addEventListener("DOMContentLoaded", boot);
               "shapes": {
                 "B1H16S4096D128": {
                   "layouts": {
-                    "contig": 10.649
+                    "contig": 0.835
                   }
                 },
                 "B8H12S1024D64": {
--- a/torch_mojo_backend/eager_kernels/aten_fast.py
+++ b/torch_mojo_backend/eager_kernels/aten_fast.py
@@ -6497,7 +6497,11 @@ def _fa4_16bit_d64_causal_inputs(
     gate and the launch cache below; f16 only needs its own RS wgmma
     emitter because the stdlib's register-operand overload is hardcoded
     bf16 (see ``fa4_wgmma_f16.mojo``). Q/K/V must all share one dtype from
-    this set -- no bf16/f16 mixing.
+    this set -- no bf16/f16 mixing. ``head_dim`` is a compile-time regime,
+    not a free runtime dimension (AGENTS.md rule 4): only 64 (GPT-2-class)
+    and 128 (Llama-class) have an instantiated kernel + exported Mojo
+    symbol (``fa4_ops.mojo``); any other head_dim declines here and falls
+    to the math decomposition.
     """
     q = _t(query)
     k = _t(key)
@@ -6525,7 +6529,13 @@ def _fa4_16bit_d64_causal_inputs(
     ):
         return None
     batch, heads, seqlen, head_dim = q._shape
-    if batch <= 0 or heads <= 0 or seqlen <= 0 or seqlen % 128 != 0 or head_dim != 64:
+    if (
+        batch <= 0
+        or heads <= 0
+        or seqlen <= 0
+        or seqlen % 128 != 0
+        or head_dim not in (64, 128)
+    ):
         return None
     if scale is not None and (
         not isinstance(scale, int | float)
@@ -6561,11 +6571,11 @@ def _fa4_strided_bthd_layout(tensor) -> bool:
         or seqlen <= 0
         or heads <= 0
         or seqlen % 128 != 0
-        or head_dim != 64
+        or head_dim not in (64, 128)
         or min(b_stride, s_stride, h_stride, d_stride) <= 0
         or d_stride != 1
-        or h_stride != 64
-        or s_stride < heads * 64
+        or h_stride != head_dim
+        or s_stride < heads * head_dim
         or b_stride != seqlen * s_stride
         or tensor._ptr % 16 != 0
     ):
@@ -6621,6 +6631,26 @@ def _fa4_prepare_qkv_bridge(q_native, k_native, v_native):
     return qkv, False
 
 
+def _fa4_symbol(
+    fa4_ops: object, direction: str, dtype: DType, head_dim: int, variant: str = ""
+) -> object:
+    """Resolve one FA4 Mojo-exported entry point by name.
+
+    Mirrors the ``flash_attention_{direction}_{bf16,f16}_d{64,128}_causal{variant}``
+    naming convention ``PyInit_fa4_ops`` registers (``fa4_ops.mojo``) --
+    allocation and symbol selection are both derived from ``head_dim`` at
+    the call site rather than hardcoded, so d64 and d128 share every
+    dispatcher below (``variant`` is ``""`` for the dense entry point,
+    ``"_strided_qkv"`` or ``"_bhsd"`` for the zero-copy ones). ``fa4_ops``
+    is typed ``object`` rather than ``ModuleType``: host-only tests pass a
+    ``SimpleNamespace`` double here instead of the real compiled extension
+    module.
+    """
+    dtype_suffix = "bf16" if dtype == DType.bfloat16 else "f16"
+    name = f"flash_attention_{direction}_{dtype_suffix}_d{head_dim}_causal{variant}"
+    return getattr(fa4_ops, name)
+
+
 def fast_fa4_16bit_d64_causal_forward(
     query,
     key,
@@ -6631,8 +6661,8 @@ def fast_fa4_16bit_d64_causal_forward(
     scale=None,
     enable_gqa=False,
 ):
-    """Run vendored FA4 (bf16 or f16) and return output/LSE plus saved
-    physical inputs.
+    """Run vendored FA4 (bf16 or f16, head_dim 64 or 128) and return
+    output/LSE plus saved physical inputs.
 
     Loading the bridge happens before materialization or allocation, so a
     packaging/compiler error cannot leave unnecessary device work queued.
@@ -6663,11 +6693,7 @@ def fast_fa4_16bit_d64_causal_forward(
         out_native = _alloc((batch, heads, seqlen, head_dim), qkv_dtype, q._device)
         logsumexp = _alloc((batch, heads, seqlen), DType.float32, q._device)
         scale_value = float(scale) if scale is not None else 1.0 / math.sqrt(head_dim)
-        forward_fn = (
-            fa4_ops.flash_attention_fwd_bf16_d64_causal_bhsd
-            if qkv_dtype == DType.bfloat16
-            else fa4_ops.flash_attention_fwd_f16_d64_causal_bhsd
-        )
+        forward_fn = _fa4_symbol(fa4_ops, "fwd", qkv_dtype, head_dim, "_bhsd")
         _device_call(
             forward_fn,
             q._ptr,
@@ -6699,11 +6725,7 @@ def fast_fa4_16bit_d64_causal_forward(
     logsumexp = _alloc((batch, heads, seqlen), DType.float32, q._device)
     scale_value = float(scale) if scale is not None else 1.0 / math.sqrt(head_dim)
     if use_strided_qkv:
-        forward_fn = (
-            fa4_ops.flash_attention_fwd_bf16_d64_causal_strided_qkv
-            if qkv_dtype == DType.bfloat16
-            else fa4_ops.flash_attention_fwd_f16_d64_causal_strided_qkv
-        )
+        forward_fn = _fa4_symbol(fa4_ops, "fwd", qkv_dtype, head_dim, "_strided_qkv")
         _device_call(
             forward_fn,
             q_native._ptr,
@@ -6722,11 +6744,7 @@ def fast_fa4_16bit_d64_causal_forward(
             keepalive=(q_native, k_native, v_native, out_native, logsumexp),
         )
     else:
-        forward_fn = (
-            fa4_ops.flash_attention_fwd_bf16_d64_causal
-            if qkv_dtype == DType.bfloat16
-            else fa4_ops.flash_attention_fwd_f16_d64_causal
-        )
+        forward_fn = _fa4_symbol(fa4_ops, "fwd", qkv_dtype, head_dim)
         _device_call(
             forward_fn,
             q_native._ptr,
@@ -6785,11 +6803,7 @@ def fast_fa4_16bit_d64_causal_backward(
         (batch * heads * seqlen_padded * head_dim,), DType.float32, q_native._device
     )
     if use_strided_qkv:
-        backward_fn = (
-            fa4_ops.flash_attention_bwd_bf16_d64_causal_strided_qkv
-            if qkv_dtype == DType.bfloat16
-            else fa4_ops.flash_attention_bwd_f16_d64_causal_strided_qkv
-        )
+        backward_fn = _fa4_symbol(fa4_ops, "bwd", qkv_dtype, head_dim, "_strided_qkv")
         _device_call(
             backward_fn,
             q_native._ptr,
@@ -6828,11 +6842,7 @@ def fast_fa4_16bit_d64_causal_backward(
             ),
         )
     else:
-        backward_fn = (
-            fa4_ops.flash_attention_bwd_bf16_d64_causal
-            if qkv_dtype == DType.bfloat16
-            else fa4_ops.flash_attention_bwd_f16_d64_causal
-        )
+        backward_fn = _fa4_symbol(fa4_ops, "bwd", qkv_dtype, head_dim)
         _device_call(
             backward_fn,
             q_native._ptr,
--- a/torch_mojo_backend/eager_kernels/aten_fast.py
+++ b/torch_mojo_backend/eager_kernels/aten_fast.py
@@ -6575,6 +6575,27 @@ def _fa4_strided_bthd_layout(tensor) -> bool:
     )
 
 
+def _fa4_bhsd_layout(tensor) -> bool:
+    """Whether the PUBLIC (B, H, S, D) tensor is eligible for the
+    BHSD-native TMA path -- no BTHD materialization needed at all.
+
+    TMA descriptor creation over the (B*H, S, D) plane view requires the
+    exact row-major (B, H, S, D) contiguity the aten op naturally
+    produces, plus a 16-byte-aligned base pointer. A sliced/offset view
+    can be fully contiguous yet still violate the latter, so both are
+    checked here -- mirroring the alignment requirement
+    ``_fa4_strided_bthd_layout`` enforces for the strided BTHD ABI.
+    """
+    return (
+        tensor is not None
+        and tensor._dtype in (DType.bfloat16, DType.float16)
+        and tensor._itemsize == tensor._dtype.size_in_bytes
+        and len(tensor._shape) == 4
+        and tensor._is_contiguous
+        and tensor._ptr % 16 == 0
+    )
+
+
 def _fa4_native_bthd(tensor):
     """Expose public BHTD storage as FA4-native BTHD, copying if required."""
     physical = fast_aten_transpose(tensor, 1, 2)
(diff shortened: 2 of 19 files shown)


torch_mojo_backend/eager_kernels/aten_fast.py
```diff
@@ -6525,7 +6529,13 @@ def _fa4_16bit_d64_causal_inputs(
     ):
         return None
     batch, heads, seqlen, head_dim = q._shape
-    if batch <= 0 or heads <= 0 or seqlen <= 0 or seqlen % 128 != 0 or head_dim != 64:
+    if (
+        batch <= 0
+        or heads <= 0
+        or seqlen <= 0
+        or seqlen % 128 != 0
+        or head_dim not in (64, 128)
```
gabrieldemarmiesse: would it be a lot of work to relax those two conditions? I feel like it's very restrictive no?

benchmarks/baselines.html
```diff
@@ -5113,12 +5132,31 @@ <h2>Worst first</h2>
           "dtypes": {
             "bf16": {
               "shapes": {
+                "B1H16S4096D128": {
+                  "layouts": {
+                    "contig": 0.592
```
gabrieldemarmiesse, line 5137: im surprised the numbers in the benchmark are so good. did we forget to return some tensors that the aten function would return by any chance? Are we ok on the accuracy part?

torch_mojo_backend/eager_kernels/aten_fast.py
gabrieldemarmiesse-ai-agent: Good instinct — they're not equally restrictive, and one of them is now removable:

**`seqlen % 128`** is a leftover from the BM=192 era. The current kernels (v2b and selfload both run BM=64, and the bhsd descriptor path zero-fills/clamps partial tiles per plane) already handle arbitrary tails — we test S=65/127/193 through the bridge directly and they pass. The gate just never caught up. I'm relaxing it on this branch now (bhsd route first, since the legacy copy routes predate the tail machinery), with aten-level tests + a benchmark spot-check for odd seqlens. Should be a small commit.

**`head_dim in (64, 128)`** is cheaper per-dim than it looks but not free: the kernel is comptime head_dim-parametric (that's how 128 landed as pure wiring), so each new dim = instantiation + SMEM/tile-fit validation + benchmark evidence, roughly what the D128 commit was. 64/128 covers GPT-2/Llama/Mistral-class models; the next real-world dims are 80 (GPT-J/NeoX), 96, and 256 (some MLA/DeepSeek shapes). Tell me which you want and I'll queue them — dims that aren't multiples of 16 would need padding logic, which is genuinely more work.

Worth noting the decline path isn't a failure either way — ineligible shapes fall to the bmm+softmax decomposition, which also got much faster this cycle (the batched-NN bmm PR feeds it).

benchmarks/baselines.html
gabrieldemarmiesse-ai-agent: Fair suspicion — checked both angles explicitly:

**Are we skipping work stock does?** No. The flash op returns a real logsumexp computed by the kernel (there's a dedicated test, `test_fa4_direct_flash_aten_returns_real_logsumexp`, that fails if we return a placeholder), and the RNG/debug outputs are legitimately empty because the eligibility gate requires `dropout_p == 0` — stock's kernel does no RNG work at dropout=0 either, so the comparison is symmetric. Both legs are measured the same way (kineto device-time sum, memcpy excluded).

**Why the D128 number (0.59 flash / 0.83 composite) is believable:** stock PyTorch's *flash* backend on Hopper is an FA2-lineage kernel that's genuinely weak at head_dim=128 — that's exactly why FlashAttention-3 was written and why torch prefers its cuDNN backend there. Our kernel is FA3-structured (wgmma+TMA, warp-specialized). The fairer 'so good?' check is the **composite** node, where stock picks its best backend (cuDNN attention): we're 0.91 (D64) / 0.83 (D128) there, not 0.5x — and it took three schedule phases to get under 1.0 against cuDNN, which is the curve you'd expect.

**Accuracy:** the harness checks sampled outputs against an fp64 reference with bf16/f16-calibrated tolerances on every shape (including tails), the aten-level tests compare against CPU at the same tolerances as the pre-existing bf16 tests, numerics policy matches stock (f32 accumulation, standard exp2 softmax with folded scale), and the selfload kernel additionally ran a 2048-launch bitwise-stability soak at three clock regimes. No tolerance was loosened anywhere in this PR.

torch_mojo_backend/eager_kernels/aten_fast.py
gabrieldemarmiesse-ai-agent: Relaxed for the no-grad BHSD forward route (d64 and d128): any `seqlen >= 1` now reaches FA4 there, since its tail machinery is already zero-fill/clamp-proven. Kept `seqlen % 128 == 0` for the legacy strided/dense routes and for anything that needs a backward pass — the bwd kernels address Q/K/V via one TMA descriptor spanning the whole flattened `(B*S, H, D)` range, so a partial last tile with `batch > 1` would silently corrupt the next batch's `dK`/`dV`, not just this one's. A training-shaped odd-seqlen call now falls through to the math-decomposition backward instead. Landed in e6a5bee with an A/B showing no cliff (S1023 vs S1024 device time within 0.2%) — details in the updated PR description.

PR #398, opened 2026-08-18
fix(fa4): round the scaled score before subtracting the rowmax

gabrieldemarmiesse-ai-agent:
## The defect

FA4's online softmax formed the probability of every score as

```mojo
P = exp2(fma(s, scale_log2, -rowmax))     # one fused multiply-add
```

The fma keeps `s * scale_log2` **exact**, but `rowmax` is the **rounded**
scaled maximum (`RN32(max_i(s_i) * scale_log2)`). Subtracting a rounded value
from an exact one leaves the winning element's exponent off zero by

```
|x - RN32(x)| <= 0.5 * ulp32(x),   x = s_max * scale_log2
```

so `p_max` is no longer exactly `1.0`. That matters because FA4 packs `P` to
16-bit for the `P@V` WGMMA while accumulating the rowsum in f32: when the
dominant entry is exactly `1.0` it is exactly representable in both, and its
packing error cancels between numerator and denominator. Once it is not, the
normalized dominant coefficient is effectively `bf16(p)/p` instead of one, and
on a saturated row that single term is the output.

The fix computes the correctly rounded product first and subtracts afterwards,
at all six sites (regular forward x2, self-load forward x2, backward
P-recompute x2):

```mojo
P = exp2(s.fma[FastMathFlag.NONE](scale_log2, 0) - rowmax)
```

**Why random-input sweeps never caught it.** The residual is an envelope that
widens with each binade of the *scaled logit*, not with saturation. At healthy
magnitudes (`|x| ~ 2e2`) half an ulp is `7.6e-6` and the error is invisible;
the 2026-08 nanoGPT run drove layer-1 logits to `|x| ~ 7e4`, where half an ulp
is `3.9e-3` and displaces the dominant probability by up to `2 ** 3.9e-3 - 1 =
0.27%`. Every existing accuracy test uses ordinary-magnitude inputs, so the
regime that triggers this is one no sweep visits. It is magnitude-triggered,
not saturation-triggered.

**The spelling is load bearing.** Written plainly as `s * scale_log2 - rowmax`
the compiler contracts it straight back into the defective fma and emits a
bit-identical kernel; `SIMD.fma` itself defaults to `FastMathFlag.CONTRACT`.
The source therefore asks for strict semantics explicitly, and a PTX guard
(below) keeps the request honest instead of trusting it.

## Causal chain, in brief

The defect was localized on a captured H100 nanoGPT-124M state that entered a
sustained gradient-norm ratchet, by faithful 600-step replays from one
ancestral checkpoint:

- **Transplant.** Replacing only `aten::_scaled_dot_product_flash_attention`
  and its backward with the CUDA ops flips the outcome: three CUDA-SDPA arms
  are clean, while the same-environment production arm and a sham arm (same
  interception plumbing, Mojo kernels retained) both ratchet.
- **Bisection.** Mojo forward + CUDA backward ratchets; CUDA forward + Mojo
  backward is clean. A forward transplant replaces output **and** LSE, so this
  establishes that the *forward-returned state* is necessary, not that output
  values alone are.
- **Proof by fix.** Replays with only the forward source expression corrected
  (and one with the full patch) are clean, 4 of 4, against controls that
  ratchet in the same environment. That is what closes the gap the bisection
  leaves.

The claim this PR makes is bounded accordingly: forward-returned state is
necessary for the ratchet in this captured trajectory, and the forward-only
source fix was sufficient in replays. It is not a claim about training runs in
general.

## Accuracy

On the captured cliff-regime tensors, saturated-row error against a CPU fp64
reference:

| kernel | saturated-row rms | saturated-row max |
|---|---|---|
| FA4, stock | 3.441e-4 | 1.574e-2 |
| CUDA flash | 2.068e-5 | 3.794e-3 |
| FA4, fixed | matches CUDA's rms/max to four significant digits | |

"Matches" means exactly that — the two printed saturated-row metrics agree.
All-row rms still differs slightly (9.188e-4 vs 9.187e-4) and no
tensor-equality test exists; this is not a claim of numerical identity with
CUDA.

The committed regression tests measure the same effect on synthetic
extreme-logit inputs across every FA4 route (see the tables in
`tests/test_eager_kernels.py`); the collapsed-regime forward rms improves by
28-51x and dQ/dK rms by 45-12800x depending on dtype and head_dim.

## Regression tests

Two layers, both new.

**Runtime, `tests/test_eager_kernels.py`** — 44 cells over
`{bf16, f16} x {d64, d128} x {gapped_qkv, bshd_view, bhsd} x {healthy,
collapsed}`, plus a self-load-sized forward case and a backward case:

- the reference is a **CPU fp64 causal decomposition**, never a vendor kernel,
  so a future flash implementation for another architecture inherits the guard
  as written;
- Q/K/V are rounded to the target 16-bit dtype *before* the fp64 reference
  consumes them, so only kernel error is measured;
- every cell **asserts its route**: the compiled FA4 bridge symbol it expects
  is spied and required to run exactly once. Without that a future
  eligibility-gate change could route the case to the accurate math
  decomposition and leave a green test that no longer tests FA4;
- the `healthy` half of every parametrization is unchanged by the fix and
  guards against paying for it in the ordinary regime;
- tolerances are **empirical fixed-seed thresholds**, not a derived format
  floor. Every stock/fixed measurement used to pick them is tabulated in the
  source next to the bounds, with the margin each bound carries (wide on rms;
  about 1.4x either side on the peak, which is a single-element statistic);
- **coverage is H100-only** — `mojo_h100` skips everything else and FA4
  eligibility is hard-gated to `sm_90a`. Nothing here establishes portability
  to another architecture.

The backward case checks dQ/dK/dV against CPU fp64 autograd. It exercises the
fixed forward and fixed backward as a **pair**, which is the only combination
that ships (the backward consumes the forward's LSE), and does not isolate the
backward edit. The backward's own justification is consistency rather than the
forward's cancellation argument: it subtracts a rounded LSE, never a rowmax, so
no exact winner-zero exists there either way — but the fixed forward now
defines its softmax over `RN32(s * scale_log2)`, and the recompute must rebuild
P from those same rounded scores or it differentiates a different softmax than
the one whose output and LSE were saved. One of the two backward sites is in
the `softcap_on` branch, which **no exported FA4 entry instantiates**: it is
changed only so the branches cannot drift, is not compiled into any shipping
kernel, and is covered by no test.

**Codegen, `tests/test_fa4_rounded_product_ptx.py`** — cross-compiles
`fa4_ops.mojo` for `sm_90a` (no GPU, never takes the GPU lock) and asserts on
the emitted PTX that **no `ex2.approx` input is produced by an `fma`**, plus
the positive counterpart that the rounded-product form is actually present.
This covers every instantiation the module compiles — both forward kernels and
the live backward P-recompute, bf16 and f16, d64 and d128 — including the ones
no runtime test can reach, and it is what makes the fix robust to a future
toolchain deciding to contract the expression again.

### Red / green

Runtime tests, stock tree (`upstream/main`, tests only):

```
22 failed, 22 passed, 1724 deselected in 8.73s
```

Every failure is a `collapsed` cell; every `healthy` cell passes. Example:

```
E  AssertionError: saturated-row rms 1.123e-03 exceeds the collapsed bound 1.500e-04
E  AssertionError: dQ rms 1.752e-01 exceeds the collapsed bound 1.000e-04
```

Runtime tests, this branch:

```
44 passed, 1726 deselected, 17 warnings in 3.94s
```

PTX guard, stock tree:

```
2 failed
E  AssertionError: an FA4 exp2 input is produced directly by an fma ...
   Offending kernels: {bwd_main_kernel...: 64, ..., fwd_fa4_kernel...: 128, ...,
                       fwd_fa4_selfload_kernel...: 128, ...}
```

PTX guard, this branch:

```
2 passed
```

Across the 16 emitted kernels this is a total discriminator: **1728 of 1728**
`ex2` sites are fma-fed before the fix and **0** after.

## Blast radius

**Assembly.** `scripts/compare_kernel_asm.py`, stock vs this branch:

| kernel dir | accelerator | result |
|---|---|---|
| `eager_flash_attention` | `sm_90a` | **16 of 24 kernels differ** — intended, see below |
| `eager_flash_attention` | `gfx942` | does not build on either side (`failed to run the pass manager for offload functions`): FA4 is Hopper-only (WGMMA/TMA), so it emits no AMD device code at all |
| `eager_flash_attention` | `apple-m4` | same, no Metal device code |
| `eager_kernels` (all 22 modules) | `sm_90a` | **0 of 3848 kernels differ**, across 269 specializations of all 22 modules |

The 16 changed sm_90a kernels are **intended and global**: this is a numerical
change to a shared source, so every bf16/f16, d64/d128, regular/self-load and
backward instantiation is supposed to move. The 8 unchanged ones are the
backward preprocess/convert kernels, which carry no `exp2`. Nothing outside
`eager_flash_attention` changed source at all — the diff touches three `.mojo`
files, all in that directory.

**Dispatch (host code) is unchanged.** No launch geometry, gate threshold or
SM-count-derived constant is touched, so the assembly diff is the whole story
here rather than half of it.

**Direct A/B microbenchmark.** Device time only, torch-profiler kernel events,
process-level ABBA ordering (fixed, stock, stock, fixed x2), core clock pinned
at 1400 MHz. Per-config burst spread was 0.0-1.2%, so these deltas are real:

| config | stock us | fixed us | ratio |
|---|---:|---:|---:|
| fwd bf16 regular d64 | 12.50 | 12.89 | 1.032 |
| fwd bf16 regular d128 | 13.43 | 13.81 | 1.028 |
| fwd bf16 selfload d64 (nanoGPT 8x12x1024) | 59.57 | 60.74 | 1.020 |
| fwd bf16 selfload d64 (8x32x256) | 21.25 | 21.54 | 1.014 |
| fwd bf16 d128 1x16x4096 | 177.84 | 185.62 | 1.044 |
| fwd f16 regular d64 | 12.44 | 12.88 | 1.035 |
| fwd f16 regular d128 | 13.47 | 13.80 | 1.025 |
| fwd f16 selfload d64 (nanoGPT) | 59.78 | 60.90 | 1.019 |
| fwd f16 selfload d64 (8x32x256) | 21.30 | 21.79 | 1.023 |
| fwd f16 d128 1x16x4096 | 181.46 | 188.08 | 1.037 |
| fwd+bwd bf16 d64 | 116.77 | 116.53 | 0.998 |
| fwd+bwd bf16 d128 | 99.34 | 99.46 | 1.001 |
| fwd+bwd f16 d64 | 116.49 | 116.41 | 0.999 |
| fwd+bwd f16 d128 | 99.87 | 99.99 | 1.001 |

So this **does** cost something: the forward is 1.4-4.4% slower, because the
P loop gains one `sub.f32` per element where the old form fused it away. A
forward+backward step is unchanged within noise, the backward being dominated
by its other work. This is a correctness fix and the cost is stated rather
than claimed away.

(A follow-up could recover it: folding `scale_log2` into Q once, before the
QK WGMMA, would remove the per-element multiply entirely and still leave a
correctly rounded score. That is a separate change with its own precision
question, and is not attempted here.)

**Benchmark suite.** `uv run pytest benchmarks/test_attention.py` — **20
passed**. Note what that guarantees: no node regressed more than the checker's
8% dead band versus the recorded baseline. It is not a base-vs-patch
measurement; the table above is.

**Neighbour tests.** `uv run pytest tests/ -k "fa4 or flash or sdpa or
attention"` — **247 passed, 8 skipped, 0 failed** in 82s (the 209 pre-existing nodes plus
the 38 added here).

## Not in this PR

A separate, **pre-existing** wrong answer exists for `S % 128 != 0` with a
strided (non-BHSD) Q/K/V input: saturated-row max error `2.758`, rms `1.006e-1`
on both the stock and fixed trees. It is unaffected by this change and is not
in the FA4 strided kernel: `aten_fast.py:6905-6911` explicitly declines FA4 for
that combination, so what is producing the wrong answer is the **fallback**
that input reaches. It needs a route-spied red test to localize it properly and
is tracked separately. Production `S=1024` and every cell added here use
`S % 128 == 0`, so no test in this PR touches it.

### DIFF
--- a/tests/test_eager_kernels.py
+++ b/tests/test_eager_kernels.py
@@ -1,7 +1,11 @@
 """Tests for the Mojo-extension fast path used by mojo eager mode."""
 
+import collections
+import contextlib
+import functools
 import math
 import weakref
+from collections.abc import Iterator
 from pathlib import Path
 from types import ModuleType, SimpleNamespace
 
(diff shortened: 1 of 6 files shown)


gabrieldemarmiesse-ai-agent:
## Full-scale training campaign: 3 seeds × 2 backends, 2.5B tokens each

The campaign this PR's body promised has finished — 6 full nanoGPT-124M runs (50,862 steps / 2.5B tokens each, 19.2h serialized on one H100), seeds 1337/1338/1339 on both backends, mojo legs loading this branch (80cfaeb) under a **CPU-only** torch 2.11.0+cpu venv.

**The fix holds at full scale, and the two backends are statistically indistinguishable.**

| seed | mojo (this PR) | cuda | mojo − cuda | destabilization |
|---|---|---|---|---|
| 1337 | 3.246508 | 3.234260 | +0.012248 | none / none |
| 1338 | 3.236997 | 3.231008 | +0.005989 | none / none |
| 1339 | 3.242762 | 3.253146 | −0.010384 | transient / **1241-step episode** |
| **mean** | **3.242089** (sd 0.0048) | **3.239471** (sd 0.0120) | **+0.002617** | |

Paired across the 3 seeds: t(df=2) = 0.388, p ≈ 0.73 — not significant, and the sign isn't even consistent (2 seeds favour cuda, 1 favours mojo). The per-seed cross-backend gap (0.0060–0.0122) is the same size as or smaller than the within-backend seed-to-seed spread (mojo 0.0095, cuda 0.0221), which is the comparison that matters. Per-run eval-sampling noise is stdev ~0.0015.

For the seed that motivated this PR, **mojo_1337 went 3.4899 → 3.2465** (cuda: 3.2401 → 3.2343). Its train-loss median now descends straight through the window that broke the original run — at step 33k, post-fix mojo 3.3385 vs original mojo 3.4320 (broke, never recovered) vs cuda 3.3429. Gradient norms in the 30k–36k window: mojo median 0.626 with **zero** rolling-median>2 excursions, cuda 0.646 with zero. The pre-fix run sustained a rolling median of 33–37 there.

A pre-flight 600-step replay from the captured pre-entrance state confirmed this branch's port matches the original fix worktree (gnorm median 0.722, val 3.3509 — the fix class, not the stock class) before the GPU-hours were spent.

### One finding worth flagging, which cuts against a simple reading of this PR

The largest gradient-norm episode of the entire campaign belongs to **cuda**, not mojo: cuda_1339 spent 1,241 consecutive steps (28604–29844) at a rolling-median gnorm peaking at 24.98, single-step max 484.5, and its val loss went *backward* across an eval interval (25k 3.4453 → 30k 3.4571). That is qualitatively the same signature as the pre-fix mojo failure, on stock CUDA with no mojo code in the process.

So this instability belongs to the **training config**, not to a backend. What the FA4 defect did was raise mojo's probability of entering it — which is exactly what the replay matrix measured (mojo re-entered ~always from the captured state, cuda ~never) and what proof-by-fix confirmed. The other distinction that matters: cuda_1339 **recovered** (finished 0.02 behind its peers), while the pre-fix mojo run ran away (0.25 behind). Post-fix, nothing ran away on either backend.

Full state (model + optimizer + RNG + data position) was captured at the entrance to the cuda episode — `postfix_runs/cuda_1339/ratchet_step2866{5,6,7,8}.pt` — for anyone who wants to attack the config-level instability (qk-layernorm, logit soft-capping, lower peak LR remain untested proposals).

Full record, including the per-leg tripwire inventory and two disclosed harness defects of my own, is in `/scratch/equiv-exp/FINDINGS.md` §11.

PR #399, opened 2026-08-19
Support conv1d (rank-3 aten::convolution) on the mojo eager device

gabrieldemarmiesse-ai-agent:
## Summary

`fast_aten_convolution` (mojo eager device) gated on rank-4 input/weight
and declined every rank-3 call unconditionally, so any `nn.Conv1d` layer
raised `NotImplementedError` on `torch.device("mojo")` in eager mode.
`nn.functional.conv1d` is `CompositeImplicitAutograd` and decomposes
straight into `aten::convolution` with a rank-3 input and 1-element
stride/padding/dilation, so this blocked **any 1-D-conv audio/signal
model** — most notably Whisper, whose audio encoder frontend is two
back-to-back `Conv1d` layers, blocked at the very first layer.
`torch.compile(backend=mojo_backend)` already supported conv1d (it
reshapes through the 2-D path in `aten_functions.py`); only the eager
fast-kernel path had the gap.

## What was broken

```
F.conv1d(x[2,3,17], w[4,3,5], padding=2)  # on mojo eager
NotImplementedError: aten::convolution is not supported by mojo eager
mode for these inputs
```

## Fix

Pure reuse, no new kernel: insert a synthetic size-1 `H` axis into the
input/weight via a zero-copy `unsqueeze` (rank 3 `(N,C,L)` /
`(out_c,c_per_group,K)` -> rank 4 `(N,C,1,L)` / `(out_c,c_per_group,1,K)`),
map the single stride/padding/dilation onto the `W` axis while pinning the
dummy `H` axis at kernel=1/stride=1/pad=0/dilation=1 (so `out_h` is always
1), recurse into the existing rank-4 im2col+GEMM conv2d path, then squeeze
the axis back out of the result (also zero-copy). Bias, stride, padding,
dilation, and groups all fall out of the existing conv2d logic unchanged.
`transposed=True` is declined exactly like the rank-4 path already
declines it (same `not transposed` guard, so it falls through to the same
clean `NotImplementedError`, never a crash).

Backward is untouched: conv1d shares `aten::convolution`'s forward
registration rather than adding a separate one, so it inherits conv2d's
existing autograd behavior (tracked/fixed independently) instead of
opening a new, conv1d-specific crash path.

## What this unblocks

Whisper's audio encoder frontend (and any other 1-D-conv model — wav2vec2-
style, WaveNet-ish) can now run its `Conv1d` layers on the mojo device in
eager mode.

## Test plan

- [x] `uv run pytest tests/test_aten_functions.py -k conv1d -v` — 8 passed
      (CallChecker pattern, `mojo:cpu` eager vs CPU reference; covers
      bias/no-bias, stride, padding, dilation, groups, f32/f16).
- [x] `flock /tmp/gpu_lock_0.lock uv run pytest tests/test_eager_kernels.py -k conv1d -v`
      — 10 passed on H100 (bias/no-bias, stride, padding, dilation, groups,
      bf16/f16, an awkward non-round length, Whisper-shaped `(80, 3000)` /
      `(384, 3000)` stride-1/stride-2 shapes, and a clean-decline check for
      `conv_transpose1d`).
- [x] `flock /tmp/gpu_lock_0.lock uv run pytest tests/test_eager_kernels.py -k conv -v`
      — 28 passed (existing conv2d coverage unaffected).
- [x] `uvx pre-commit run --all-files` — clean.

### DIFF
--- a/tests/test_aten_functions.py
+++ b/tests/test_aten_functions.py
@@ -941,6 +941,70 @@ def test_aten_bitwise_xor_broadcasting(conf: Conf):
     check_outputs(fn, conf, [x, y])
 
 
+@pytest.mark.parametrize(
+    "in_c,out_c,length,k,stride,padding,dilation,groups",
+    [
+        (3, 8, 17, 3, 1, 0, 1, 1),  # basic, awkward/non-round length
+        (8, 8, 17, 3, 2, 1, 1, 1),  # stride
+        (8, 12, 25, 3, 1, 2, 2, 1),  # dilation
+        (8, 12, 25, 3, 1, 1, 1, 2),  # groups
+        (80, 16, 40, 3, 1, 1, 1, 1),  # Whisper-shaped channel counts
+    ],
+)
+def test_aten_conv1d(
+    conf: Conf,
+    call_checker: CallChecker,
+    in_c: int,
+    out_c: int,
+    length: int,
+    k: int,
+    stride: int,
+    padding: int,
+    dilation: int,
+    groups: int,
+):
+    """aten::convolution with a rank-3 (conv1d) input.
+
+    Same overload as conv2d; mojo eager reuses the rank-4 im2col+GEMM path by
+    inserting a synthetic size-1 H axis (see `fast_aten_convolution`).
+    """
+    call_checker.register(aten_functions.aten_convolution)
+
+    def fn(x, w, b):
+        return torch.nn.functional.conv1d(
+            x, w, b, stride=stride, padding=padding, dilation=dilation, groups=groups
+        )
+
+    x = torch.randn(2, in_c, length)
+    w = torch.randn(out_c, in_c // groups, k)
+    b = torch.randn(out_c)
+    check_outputs(fn, conf, [x, w, b], atol=1e-4, rtol=1e-4)
+
+
+def test_aten_conv1d_no_bias(conf: Conf, call_checker: CallChecker):
+    call_checker.register(aten_functions.aten_convolution)
+
+    def fn(x, w):
+        return torch.nn.functional.conv1d(x, w, padding=1)
+
+    x = torch.randn(2, 4, 13)
+    w = torch.randn(6, 4, 3)
+    check_outputs(fn, conf, [x, w], atol=1e-4, rtol=1e-4)
+
+
+@pytest.mark.parametrize("dtype", [torch.float32, torch.float16])
+def test_aten_conv1d_dtypes(conf: Conf, call_checker: CallChecker, dtype: torch.dtype):
+    call_checker.register(aten_functions.aten_convolution)
+
+    def fn(x, w, b):
+        return torch.nn.functional.conv1d(x, w, b, stride=2, padding=1)
+
+    x = torch.randn(2, 8, 21, dtype=dtype)
+    w = torch.randn(12, 8, 3, dtype=dtype)
+    b = torch.randn(12, dtype=dtype)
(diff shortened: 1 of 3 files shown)


gabrieldemarmiesse:
Fix conflicts

gabrieldemarmiesse-ai-agent:
Rebased onto latest `main` (past #401 and #402). One conflict, in `tests/test_eager_kernels.py`: #401 inserted `test_fast_conv2d_refuses_autograd_in_the_forward` at the same spot (right after `test_fast_conv2d_batched_falls_back_correctly`) that this PR's conv1d tests were added — resolved by keeping both blocks (upstream's preflight test first, then the conv1d tests), no logic changes needed on either side.

Worth noting: #401 also changed `aten::convolution`'s registration from `_register_fast` to the new `mojo_device_convolution` preflight wrapper (in `mojo_device_aten_ops.py`, which this PR never touched), but that wrapper calls straight through to `fast_aten_convolution` — the function this PR extends — so conv1d now gets the clean forward-time backward-preflight refusal for free, exactly as this PR's description anticipated (conv1d shares conv2d's forward registration and inherits its autograd story).

Re-verified after rebase:
- `uv run pytest tests/test_aten_functions.py -k conv1d -v` — 8 passed
- `flock /tmp/gpu_lock_0.lock uv run pytest tests/test_eager_kernels.py -k conv -v` — 29 passed (incl. upstream's new `test_fast_conv2d_refuses_autograd_in_the_forward`)
- `flock /tmp/gpu_lock_0.lock uv run pytest tests/test_eager_kernels.py -k "div or autograd_preflight" -q` — 68 passed (sanity on the other files #401/#402 touched)
- `uvx pre-commit run --all-files` — clean

Force-pushed as `0403aeb`. PR now shows `mergeable: MERGEABLE`.

gabrieldemarmiesse-ai-agent:
Fixed both blocking findings from the adversarial review:

**1. Invalid-groups silent wrong result (blocking).** `fast_aten_convolution`'s rank-4 path validated only `c_per_group * groups == in_channels`, never `out_channels % groups`. Added that guard next to it — fixes both ranks with one condition since conv1d recurses into the rank-4 path. Verified the exact repro:

```python
F.conv1d(torch.randn(1, 4, 9), torch.randn(5, 2, 3), groups=2)
```

now raises `NotImplementedError` on mojo (stock raises `RuntimeError`) instead of returning a `(1, 5, 7)` tensor with an unwritten channel. Added `test_fast_conv1d_invalid_groups_declines_instead_of_partial_output` and the rank-4 twin as regression tests.

**2. Native backward crash (blocking).** Verified directly on H100 with the review's repro:

```python
x = torch.randn(1, 2, 9).to("mojo:0").requires_grad_()
w = torch.randn(3, 2, 3).to("mojo:0")
F.conv1d(x, w, padding=1).sum().backward()
```

This branch's rebase onto `main` (past #401) already closes this: it now raises `NotImplementedError: aten::convolution would record an autograd node whose backward needs aten::convolution_backward, which mojo eager mode does not implement...` instead of aborting. #401's preflight wrapper (`mojo_device_convolution`) calls straight into `fast_aten_convolution`, so conv1d gets it for free via the shared `aten::convolution` registration. Added `test_fast_conv1d_refuses_autograd_in_the_forward` (mirroring the existing conv2d test) to pin this specifically for rank-3, since it wasn't otherwise tested on this branch.

32/32 conv-related GPU tests passing (`flock /tmp/gpu_lock_0.lock uv run pytest tests/test_eager_kernels.py -k conv -v`), `pre-commit` clean. Pushed as `b612a99`.

PR #400, opened 2026-08-19
Support enable_gqa=True in eager SDPA (grouped-query attention)

gabrieldemarmiesse-ai-agent:
## What was broken

`torch.nn.functional.scaled_dot_product_attention(q, k, v, is_causal=True, enable_gqa=True)` raised `NotImplementedError` on the `mojo` eager device whenever K/V carried fewer heads than Q. Grouped-query attention had **no** eager implementation at all:

- `_fa4_16bit_d64_causal_inputs` and `_fused_fa_inputs` both check `or enable_gqa` and decline outright,
- the masked branch of `fast_aten_scaled_dot_product_attention` declined `enable_gqa` too,
- and its bmm decomposition gate required `tuple(q._shape[:2]) == tuple(k._shape[:2])`, i.e. equal head counts.

## What it unblocks

HF `transformers`' `sdpa_attention_forward` stops calling `repeat_kv` itself and passes `enable_gqa=True` whenever there is no padding `attention_mask` — that is **every ordinary causal forward/generate step** of Llama-3, Mistral, Qwen2, Gemma2 and Phi-3 on the mojo device.

## The fix

Expand K/V's head axis to Q's **once, in the entry points**, then let every existing equal-head route run unchanged — no new kernel, no new shape regime.

- `aten_fast._expand_kv_heads_for_gqa` is `repeat_kv` over `TorchMojoTensor`. `kv_heads == 1` (MQA) is a genuine zero-copy stride-0 broadcast that the decode kernel reads through; other ratios materialize, because the `(kv_heads, n_rep)` merge after a broadcast expand is not expressible as a view — the same reason HF's own `repeat_kv` materializes.
- Expanding in the entry point rather than inside the decomposition means a GQA prefill now reaches **FA4** instead of the O(n²) math path. Measured: a Llama-esque 32/8-head, head_dim 128, causal bf16 call lands on `flash_attention_fwd_bf16_d128_causal_bhsd` (pinned by a test).
- The **backward-needing** route in `mojo_device_autograd` expands with ordinary differentiable ops instead, so PyTorch's own `ExpandBackward` reduces `grad_key`/`grad_value` back to the leaves' `(b, kv_heads, s, d)` shape with no custom backward code. It runs **strictly after** the native-flash dispatch gate: `_scaled_dot_product_flash_attention`'s backward is a native node, and a decline inside a native backward node aborts the process (`std::terminate`) instead of raising, so a GQA triple must never become "eligible" for it by being expanded first. The forward-only route is deliberately left to `aten_fast` so MQA decode keeps its zero-copy broadcast.
- `aten::_scaled_dot_product_attention_math` had the identical decline and gets the identical treatment.

The `torch.compile` backend already handled `enable_gqa` correctly; `aten_functions.py` is unchanged.

## Test evidence

37 new tests in `tests/test_eager_kernels.py`, all passing (H100):

- forward parity vs a CPU fp32 `repeat_kv` reference across 32:8 and 8:1 (MQA) ratios × bf16/f16 × head_dim 64/128 × causal/non-causal (16 cases),
- grad-enabled-but-graph-free fp32 forward (math decomposition), masked (`attn_mask` + GQA, which used to decline), strided Q/K/V taken as transposed views of one packed projection,
- backward parity with grads asserted to land in the **original** `(b, kv_heads, s, d)` shape, causal and non-causal,
- the HF `sdpa_attention_forward` call shape at Llama-3-8B's 32/8 heads, head_dim 128 — prefill and one decode step,
- unit tests that the MQA expansion is zero-copy (same `_ptr`, same holder, stride 0) and that the materialized expansion equals `repeat_kv`,
- route pins: a GQA prefill reaches `flash_attention_fwd_bf16_d{64,128}_causal_bhsd`; a backward-needing GQA call whose *expanded* form would be FA4-eligible does **not** get `ScaledDotProductFlashAttentionBackward0`; a no-grad MQA call never touches the differentiable `repeat_kv`,
- indivisible head counts (6:4, 2:3) still raise `NotImplementedError` cleanly, and `enable_gqa=True` with equal head counts is bit-identical to `enable_gqa=False`.

Regression: `pytest tests/test_eager_kernels.py -k "sdpa or attention or fa4 or flash"` → **164 passed, 4 skipped**; `tests/test_sdpa_host_wiring.py tests/test_fa4_host_wiring.py tests/test_mojo_autocast.py` → **27 passed**; `tests/test_high_level_ops.py -k "sdpa or attention"` → **26 passed**; `benchmarks/test_coverage.py` → 2 passed. `uvx pre-commit run --all-files` clean.

No Mojo kernel was touched, so no cross-architecture assembly check is owed: this is Python-level view/copy composition over already-shipped kernels.

### DIFF
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -8677,6 +8677,16 @@ document.addEventListener("DOMContentLoaded", boot);
                     "contig": 0.832
                   }
                 },
+                "B1H32KV8S4096D128": {
+                  "layouts": {
+                    "contig": 4.032
+                  }
+                },
+                "B1H8KV1S4096D128": {
+                  "layouts": {
+                    "contig": 0.923
+                  }
+                },
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.912
@@ -8689,6 +8699,16 @@ document.addEventListener("DOMContentLoaded", boot);
                 "B1H16S4096D128": {
                   "layouts": {
                     "contig": 0.835
+                  }
+                },
+                "B1H32KV8S4096D128": {
+                  "layouts": {
+                    "contig": 3.988
+                  }
+                },
+                "B1H8KV1S4096D128": {
+                  "layouts": {
+                    "contig": 0.929
                   }
                 },
                 "B8H12S1024D64": {
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -8639,6 +8639,16 @@ document.addEventListener("DOMContentLoaded", boot);
                     "contig": 0.832
                   }
                 },
+                "B1H32KV8S4096D128": {
+                  "layouts": {
+                    "contig": 4.032
+                  }
+                },
+                "B1H8KV1S4096D128": {
+                  "layouts": {
+                    "contig": 0.923
+                  }
+                },
                 "B8H12S1024D64": {
                   "layouts": {
                     "contig": 0.912
@@ -8651,6 +8661,16 @@ document.addEventListener("DOMContentLoaded", boot);
                 "B1H16S4096D128": {
                   "layouts": {
                     "contig": 0.835
+                  }
+                },
+                "B1H32KV8S4096D128": {
+                  "layouts": {
+                    "contig": 3.988
+                  }
+                },
+                "B1H8KV1S4096D128": {
+                  "layouts": {
+                    "contig": 0.929
                   }
                 },
                 "B8H12S1024D64": {
--- a/benchmarks/test_attention.py
+++ b/benchmarks/test_attention.py
@@ -6,6 +6,11 @@ entry point is pinned — the public F.scaled_dot_product_attention node
 additionally measures whatever backend selection stock PyTorch performs.
 Shapes are (batch, heads, seq, head_dim): a GPT-2-like block and a long
 single-sequence decode-prefill regime.  All cases are causal.
+
+GQA_SHAPES cover the enable_gqa=True regime (see fix/gqa-sdpa-enable-gqa):
+Q and K/V carry different head counts, so they get their own shape dict
+keyed as B{batch}H{q_heads}KV{kv_heads}S{seq}D{head_dim} — the KV token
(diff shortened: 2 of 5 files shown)


gabrieldemarmiesse:
It's missing a new benchmarking entry no?

gabrieldemarmiesse-ai-agent:
Good catch — added two GQA-shaped nodes to `benchmarks/test_attention.py` for `aten::scaled_dot_product_attention` under `enable_gqa=True` (commit 972c2ad), following the file's `B{batch}H{heads}S{seq}D{head_dim}` convention with a `KV{kv_heads}` token so the ratio being measured is legible in the shape id itself:

- `B1H32KV8S4096D128` — Llama-3-8B's own head ratio (32 query heads / 8 KV heads, head_dim 128), the regime that materializes K/V (this backend's `repeat_kv`-style expand+copy for any GQA ratio other than pure MQA).
- `B1H8KV1S4096D128` — an MQA extreme (kv_heads=1), the zero-copy stride-0 broadcast case.

Measured on H100 (bf16/f16, S=4096), recorded via `--update-baselines`:

| shape | bf16 ratio | f16 ratio |
|---|---|---|
| B1H32KV8S4096D128 | 4.032 | 3.988 |
| B1H8KV1S4096D128 | 0.923 | 0.929 |

Honest number as requested: the materializing 32:8 case is **~4x slower than stock PyTorch**, well outside the 1.10 bar. The MQA (8:1) case is actually faster than stock (~0.92-0.93x), matching the zero-copy design for `kv_heads=1`.

The 4x gap on the materializing path is a real, known follow-up — this PR's scope was making GQA *correct* on the mojo device via pure reuse (expand + existing SDPA kernels), not a dedicated GQA-aware fused kernel. A proper fix needs a kernel that reads K/V through a broadcast head-stride directly (no materialization), which is separate kernel work, not a registration fix. Filing this as a follow-up rather than blocking on it here, since today the alternative is `NotImplementedError` for every GQA model.

`benchmarks/test_coverage.py`'s two reconciliations both pass; the baselines.html diff is exactly these 4 new entries, nothing else touched.

PR #402, opened 2026-08-19
Support div rounding_mode="floor"/"trunc" on the mojo eager device

gabrieldemarmiesse-ai-agent:
## Summary

`aten::div.Tensor_mode` / `aten::div.out_mode` had **no eager registration
at all**, so `torch.div(x, y, rounding_mode="floor"/"trunc")` (and Python's
`//` operator on mojo tensors, which routes through the same overload)
raised the raw dispatcher error before ever reaching our code:

```
Could not run 'aten::div.out_mode' with arguments from the 'mojo' backend...
```

`fast_aten_div` did have a `if rounding_mode is not None: return
NOT_HANDLED` guard, but it was unreachable dead code for this overload —
the real blocker was the missing registration.

## Fix

- Registered `aten::div.Tensor_mode` / `aten::div.out_mode` (mirroring the
  existing `div.Tensor` / `div.out` pair). `div.Scalar_mode` needs no
  separate registration: it's `CompositeExplicitAutograd` and decomposes
  into `Tensor_mode`, verified empirically to reach `fast_aten_div` with
  the Python scalar intact (the same mechanism plain `div.Scalar` already
  relies on today).
- Wired `fast_aten_div`'s `rounding_mode` branch to two broadcast-strided
  kernels: `"floor"` reuses the already-shipped `FloorDivSpec`
  (`aten::floor_divide`'s kernel, untouched); `"trunc"` is one new kernel,
  `BOP_TRUNCDIV` in `logic_ops.mojo`, in the same elementwise-op family.

## The actual work: matching torch's sign and precision semantics exactly

Floor and trunc division only disagree when the operands have **opposite
signs and the division is inexact**. Integer trunc-division corrects
Mojo's native floor-division (`//`) by one toward zero in exactly that
case.

For narrow floats (bf16/fp16), floor and trunc need **opposite** precision
strategies to match torch bit-for-bit:
- `floor_divide` upcasts to fp32 (torch's own fmod-based algorithm is
  effectively exact there — this was already the case before this PR).
- torch's tensor-tensor **trunc** kernel does *not* upcast for the general
  case (only its separate scalar-operand fast path does — checked directly
  against `aten/src/ATen/native/cpu/BinaryOpsKernel.cpp`), so
  `BOP_TRUNCDIV` intentionally stays at native precision. Upcasting there
  would be "more correct" than torch itself and silently diverge from it —
  confirmed by a failing test before this fix (a quotient of ~5.985 rounds
  to 6.0 in bf16 before truncation, and that IS torch's real output here).

## What this unblocks

Index-arithmetic patterns that use `torch.div(..., rounding_mode=...)` or
`//` on mojo tensors — used in Swin-style window partitioning, position-id
math, and any code using Python's `//` operator on a mojo tensor.

## Test plan

- [x] `uv run pytest tests/test_aten_functions.py -k div_rounding -v` — 11
      passed (CallChecker pattern, `mojo:cpu` eager vs CPU reference; incl.
      a dedicated floor-vs-trunc-disagree case).
- [x] `flock /tmp/gpu_lock_0.lock uv run pytest tests/test_eager_kernels.py -k div -v`
      — 44 passed on H100: every sign combination for
      f32/f64/i32/i64/uint8, negative Python-scalar divisors, the `out=`
      variant, and the bf16/fp16 narrow-float boundary case pinned against
      torch's actual (non-upcast) trunc output. No regression in existing
      `floor_divide`/`remainder` coverage.
- [x] `uv run pytest benchmarks/test_coverage.py` — passed (`div.out_mode`
      classified alongside `div.out`; a new `test_div_trunc_mode` benchmark
      covers `div.Tensor_mode`'s new `TruncDivSpec` kernel).
- [x] `uvx pre-commit run --all-files` — clean.


benchmarks/test_binary.py
```diff
@@ -277,6 +279,37 @@ def test_floor_divide(
         )
 
 
+@pytest.mark.parametrize("dtype_id", ("f32",))
+@pytest.mark.parametrize("layout", ("Scalar", "Tensor"))
+@pytest.mark.parametrize("shape_id", SHAPES)
+def test_div_trunc_mode(
```
gabrieldemarmiesse, line 285: You added a new bench but i don't see the html-json updated here

gabrieldemarmiesse-ai-agent: Good catch, fixed. Recorded the missing baseline entries via a scoped `--update-baselines` run (bare, no outer flock, since `benchmarks/` self-locks):

```
uv run pytest benchmarks/test_binary.py -k "test_div_trunc_mode" --update-baselines
```

4 new leaves recorded, exactly matching the new node, nothing else touched:

- `div_trunc_mode/f32/C_4096x4096/Scalar`: ratio 1.024 (ours/stock: 78.3us / 76.6us)
- `div_trunc_mode/f32/C_4096x4096/Tensor`: ratio 1.003 (110.3us / 109.9us)
- `div_trunc_mode/f32/A_357x789/Scalar`: ratio 1.883 (3.4us / 1.8us)
- `div_trunc_mode/f32/A_357x789/Tensor`: ratio 1.041 (2.4us / 2.3us)

The A_357x789/Scalar ratio looks high but it's a tiny-absolute-time (microsecond-scale) launch-bound artifact of that awkward shape's scalar layout, the same regime `test_floor_divide`'s own Scalar/Tensor split exists to isolate — not a real regression, and there's no prior baseline to regress against since this is a new kernel (TruncDivSpec).

Verified `benchmarks/test_coverage.py` still passes (baseline still addressable by its test node). Pushed as `df36559`.

gabrieldemarmiesse-ai-agent: Confirmed: the merge commit (2612fac) includes the baseline commit (df36559) — `benchmarks/baselines.html` shows the 4 `div_trunc_mode` entries in the merged diff on main. Nothing further needed here.

PR #404, opened 2026-08-19
Add gather / index_select / scatter_add / index_put to eager mode

gabrieldemarmiesse-ai-agent:
Adds eager-mode (`.to("mojo")`) and torch.compile-backend support for the
index-op family: `gather`, `index_select`, `scatter_add`, `index_put` — plus
`index_add`, which is not optional once `index_select` exists (see *Autograd*
below) — and `take_along_dim`, which is a composite over `gather`.

Every one of them was verified missing first, on a real mojo device:

```
[MISSING] gather              Could not run 'aten::gather.out'
[MISSING] index_select        Could not run 'aten::index_select'
[MISSING] scatter_add         Could not run 'aten::scatter_add.out'
[MISSING] index_put           Could not run 'aten::_index_put_impl_'
[MISSING] index_add           Could not run 'aten::index_add.out'
[MISSING] take_along_dim      Could not run 'aten::gather.out'
```

## Two kernels, not five

All five ops are one index-space walk with different stride vectors, so they
are two new kernels next to the existing `GatherRows`/`ScatterDim` in
`data_movement_ops.mojo` — `GatherDim` reads, `ScatterAddDim` accumulates —
and what distinguishes the ops is only which stride vector is real and which
is a broadcast:

| op | index space | index strides |
|---|---|---|
| `gather` | `index.shape` | real |
| `index_select` | `out.shape` | 0 except along `dim` |
| `scatter_add` | `index.shape` | real, atomic add |
| `index_add` | `source.shape` | 0 except along `dim`, atomic add |
| `index_put` | `values.shape` | dim 0, `(1, 0, ...)` |

`index_select` on dim 0 of a contiguous input reuses the **existing**
`GatherRows` kernel unchanged (one div/mod per element instead of three), and
`index_put(accumulate=False)` reuses the **existing** `ScatterDim`. A
contiguous operand collapses to a rank-3 `(outer, n, inner)` problem whatever
its rank, so `index_select`/`index_add` have no rank cap at all; the strided
routes keep the family's rank-4 cap.

## Semantics matched to torch, deliberately

- **No bounds checking, no negative-index wrap** for
  gather/index_select/scatter_add/index_add. That is torch's own contract:
  its CPU kernels raise before launching, but the CUDA kernels only
  device-assert (`ScatterGatherKernel.cu`, `Indexing.cu`), so an
  out-of-range or negative index is undefined on an accelerator there too.
- **`index_put` DOES wrap negative indices**, because advanced indexing
  defines them. The wrap reuses the `remainder` kernel on the index tensor,
  exactly as ATen's own `wrapIndexOnce` does.
- **`scatter_add` uses atomics and is not bitwise reproducible** for float
  dtypes — neither is torch's, whose default CUDA kernel is
  `ReduceAdd`→`fastAtomicAdd`. `torch.use_deterministic_algorithms(True)` is
  not honored (torch switches to a sort-based reduction there; we do not).
  The deterministic scheme that would work here is written down next to the
  kernel: a scatter along `dim` can only collide between entries agreeing on
  every *other* coordinate, so one thread per `dim`-column is race-free with
  no atomics for every dtype and backend — but it yields only
  `total/dims[dim]` threads (64 for a `(65536, 64)` index on dim 0), so it
  needs a runtime dispatch rather than replacing this kernel.
- **`x[mask] = scalar`** does not reach a bool-mask kernel at all: ATen's own
  `_index_put_impl_` re-routes exactly that shape to `masked_fill_`
  (`canDispatchToMaskedFill`), and so does this. A mask index covers the
  LEADING dims while broadcasting is right-aligned, so the mask is padded
  with trailing 1s first — without that, a shape-`(4,)` mask on a `(4, 4)`
  tensor fills the wrong axis, which only a square tensor exposes
  (`test_aten_index_put_bool_row_mask_scalar`).
- `scatter_add`/`index_add`/`index_put(accumulate=True)` decline
  int8/int16/uint8/bool: there is no sub-32-bit atomic RMW on the GPU
  targets, so those are refused in Python rather than sent to the compiler as
  an unsatisfiable specialization.

## Why the functional overloads are registered next to the `.out` ones

`gather`, `scatter_add` and `index_add` are `structured_delegate` ops, so
registering only `aten::gather.out` looks sufficient. It is not: PyTorch's
generated wrapper allocates the `out` tensor itself on the **current** device
rather than the input's, because the PrivateUse1 device guard is a no-op here
(there is no C++ extension to register a real one). On this box that hands
`gather.out` an `out` on `mojo:0` while the inputs are on `mojo:1`, and
nothing downstream can recover. A backend kernel on the functional overload
takes precedence over the wrapper, allocates the output where the inputs are,
and saves an allocation and a dispatch.

## Autograd

The family's derivative formulas call each other
(`gather`→`scatter_add_`, `index_select`→`index_add_`,
`scatter_add`→`gather`, `index_put`/`index`→`index_put`/`index`), and a raise
from a native backward node on this device is `std::terminate`, not an
exception — so a half-implemented family is worse than none. All eight
forward/backward pairs now run:

```
[ok] gather  [ok] gather_d0  [ok] index_select  [ok] index_select_d1
[ok] scatter_add  [ok] index_add  [ok] index_put  [ok] index_put_acc
```

This also **fixes an existing abort**: `x[idx].sum().backward()`
(`aten::index.Tensor`'s backward, which is
`zeros.index_put_(indices, grad, true)`) killed the process with
`terminate called without an active exception` on `main`, on both devices,
because `_index_put_impl_` did not exist. It now returns a gradient.

One pre-existing gap is documented rather than fixed: on a mojo device that
is not the *current* one, `index_put(accumulate=False)`'s backward still
aborts, because its formula's `zeros_like(values)` is allocated by ATen on
the current device and arrives on the wrong one. Every C++ derivative that
allocates without naming a device has that problem; the fix belongs in the
device guard.

## Verification

- **Unit tests** in `tests/test_aten_functions.py`, parametrized over
  dtype × rank × which dim is indexed (0 has a dedicated fast path, the last
  is the contiguous-inner regime, a negative dim exercises normalization),
  plus strided inputs, int32 indices, duplicate indices for the accumulating
  ops, negative indices for `index_put`, broadcast `values`, and the two
  bool-mask forms.
- **Cross-architecture assembly**, `scripts/compare_kernel_asm.py`, both
  `sm_90a` and `gfx942`: **0 kernels changed, +4 new** on each. The Mojo diff
  is purely additive — no existing line was touched.
  Getting there is worth recording: `ScatterAddDim` started as `ScatterDim`
  parametrized on the reduction, and *every* way of writing that (a
  `comptime if` around the value/src branch, one inside each arm, an
  `@always_inline` store helper called from both) moved the existing kernel's
  assembly by ±3 predicates and ±25 registers, while a bare unused comptime
  parameter changed nothing. Metaprogramming that rewrites the existing
  kernel is not a refactor, so the reduction got its own (leaner — no
  scalar-value overload) body.
- **Conformance** (`conformance/known_unsupported.py` is xfail-strict, so a
  now-passing declared node fails the suite): regenerated for these
  operators. `take_along_dim`'s entry is gone — it passes end to end now.
  The `gather`/`index_select`/`scatter_add`/`index_add`/`index_put` entries
  stay, and the reason is worth knowing: their OpInfo corpora also contain
  0-d samples and rank-5 samples, both of which decline here (rank-0 is a
  small fix; rank > 5 for gather/scatter_add needs a rank-8 kernel, since
  a full-rank index cannot be collapsed). 33 nodes xfail, 0 fail.
- **Benchmarks** in `benchmarks/test_embedding.py` for all five ops, with
  baselines recorded on this H100 (device time only, ABBA-interleaved,
  reproducible to ±0.6% and matching an independent earlier run within 1%);
  `benchmarks/test_coverage.py` passes, so every newly registered overload is
  either benchmarked or carries a written skip reason.

  | op | shape | bf16 | f32 |
  |---|---|---|---|
  | `gather` | 262144×64, dim 1 | 2.70 | 2.19 |
  | `gather` | 4096×4096, dim 0 | 1.13 | 1.08 |
  | `index_select` | 262144×64, 1048576 rows, dim 0 | **0.49** | **0.52** |
  | `index_select` | 4096×4096, 8192, dim 1 | 4.31 | 2.89 |
  | `scatter_add` | 262144×64, dim 1 | 2.28 | 1.52 |
  | `scatter_add` | 4096×4096, dim 0 | 1.25 | 1.05 |
  | `index_add` | 262144×64, 1048576, dim 0 | 1.48 | 1.52 |
  | `index_add` | 4096×4096, 8192, dim 1 | 2.35 | 1.03 |
  | `index_put` accumulate | 262144×64, 65536 rows | **0.90** | **0.84** |
  | `index_put` set | 262144×64, 65536 rows | 3.04 | 2.97 |

  (ratio = ours/stock CUDA device time; lower is better.) This is an
  op-support change, not an optimization one, so these are recorded as
  honest first-implementation baselines rather than tuned numbers — but two
  of them point somewhere specific, and they are the useful part:

  - **The dim-0 routes are at or ahead of stock** (`index_select` 0.49,
    `index_put` accumulate 0.90, `gather`/`scatter_add` dim 0 within 5-25%),
    because those are whole-row moves and the coordinate math amortizes.
  - **The inner-dim routes are 2-4× off**, and the cause is index traffic,
    not payload: at `(4096, 4096)` selecting 8192 columns, every one of the
    33.5M threads reloads its own int64 index element, i.e. 268 MB of index
    reads against 67 MB of payload. Stock's kernel stages indices per block.
    Fixing that is a tiling change, i.e. its own engagement.
  - **`index_put` with `accumulate=False` is 3× off while `accumulate=True`
    is at parity** — and the slow one is the path that reuses the
    PRE-EXISTING `ScatterDim`, while the fast one is the new
    `ScatterAddDim`. A row-granular scatter (the write mirror of
    `GatherRows`, which is what this call actually is) is the obvious next
    step, and it would speed up `aten::scatter.src`/`scatter.value` too.
- **The torch.compile backend** side (`aten_functions.py`) is exercised
  separately, since the `conf` fixture only runs eager: all 12 shapes/ops
  check out against CPU eager through `torch.compile(backend=mojo_backend)`.
  It uses the GPU-capable `_nd` scatters rather than `F.scatter_add`, whose
  wrapper has no GPU kernel and inserts a silent host round trip per call.

### DIFF
--- a/benchmarks/baselines.html
+++ b/benchmarks/baselines.html
@@ -4916,6 +4916,38 @@ document.addEventListener("DOMContentLoaded", boot);
             }
           }
         },
+        "_index_put_impl_": {
+          "dtypes": {
+            "bf16": {
+              "shapes": {
+                "R_262144x64_S65536_acc": {
+                  "layouts": {
+                    "contig": 0.899
+                  }
+                },
+                "R_262144x64_S65536_set": {
+                  "layouts": {
+                    "contig": 3.041
+                  }
+                }
+              }
+            },
+            "f32": {
+              "shapes": {
+                "R_262144x64_S65536_acc": {
+                  "layouts": {
+                    "contig": 0.837
+                  }
+                },
+                "R_262144x64_S65536_set": {
+                  "layouts": {
+                    "contig": 2.957
+                  }
+                }
+              }
+            }
+          }
+        },
         "_log_softmax": {
           "dtypes": {
             "bf16": {
@@ -6812,6 +6844,38 @@ document.addEventListener("DOMContentLoaded", boot);
             }
           }
         },
+        "gather": {
+          "dtypes": {
+            "bf16": {
+              "shapes": {
+                "R_262144x64_D1": {
+                  "layouts": {
+                    "contig": 2.677
+                  }
+                },
+                "R_4096x4096_D0": {
+                  "layouts": {
+                    "contig": 1.126
+                  }
+                }
+              }
+            },
+            "f32": {
+              "shapes": {
+                "R_262144x64_D1": {
+                  "layouts": {
+                    "contig": 2.163
+                  }
+                },
+                "R_4096x4096_D0": {
+                  "layouts": {
+                    "contig": 1.073
+                  }
+                }
+              }
+            }
+          }
+        },
         "ge": {
           "dtypes": {
             "f32": {
@@ -6954,6 +7018,70 @@ document.addEventListener("DOMContentLoaded", boot);
                 "R_262144x64_I1048576": {
                   "layouts": {
                     "contig": 0.522
+                  }
+                }
+              }
+            }
+          }
+        },
+        "index_add": {
+          "dtypes": {
+            "bf16": {
+              "shapes": {
+                "R_262144x64_S1048576_D0": {
+                  "layouts": {
+                    "contig": 1.476
+                  }
+                },
+                "R_4096x4096_S8192_D1": {
+                  "layouts": {
+                    "contig": 2.338
+                  }
+                }
+              }
+            },
+            "f32": {
+              "shapes": {
+                "R_262144x64_S1048576_D0": {
+                  "layouts": {
+                    "contig": 1.517
+                  }
+                },
+                "R_4096x4096_S8192_D1": {
+                  "layouts": {
+                    "contig": 1.013
+                  }
+                }
+              }
+            }
+          }
+        },
+        "index_select": {
+          "dtypes": {
+            "bf16": {
+              "shapes": {
+                "R_262144x64_S1048576_D0": {
+                  "layouts": {
+                    "contig": 0.487
+                  }
+                },
+                "R_4096x4096_S8192_D1": {
+                  "layouts": {
+                    "contig": 4.298
+                  }
+                }
+              }
+            },
+            "f32": {
+              "shapes": {
+                "R_262144x64_S1048576_D0": {
+                  "layouts": {
+                    "contig": 0.52
+                  }
+                },
+                "R_4096x4096_S8192_D1": {
+                  "layouts": {
+                    "contig": 2.855
                   }
                 }
               }
@@ -8758,6 +8886,38 @@ document.addEventListener("DOMContentLoaded", boot);
                 "R_262144x64_S65536": {
                   "layouts": {
                     "contig": 1.183
+                  }
+                }
+              }
+            }
+          }
+        },
+        "scatter_add": {
+          "dtypes": {
+            "bf16": {
+              "shapes": {
+                "R_262144x64_D1": {
+                  "layouts": {
+                    "contig": 2.247
+                  }
+                },
+                "R_4096x4096_D0": {
+                  "layouts": {
+                    "contig": 1.247
+                  }
+                }
+              }
+            },
+            "f32": {
+              "shapes": {
+                "R_262144x64_D1": {
+                  "layouts": {
+                    "contig": 1.508
+                  }
+                },
+                "R_4096x4096_D0": {
+                  "layouts": {
+                    "contig": 1.048
                   }
                 }
               }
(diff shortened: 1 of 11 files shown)


gabrieldemarmiesse:
Fix the conflicts

PR #431, opened 2026-09-01
Add side device streams and a cross-stream free fence

gabrieldemarmiesse-ai-agent:
Side streams are extra device streams next to the mojo device's default stream — which keeps carrying every kernel, every copy and every stream-ordered free — for work that should overlap it: NCCL collectives in the stacked distributed PR, prefetch copies later.

The load-bearing half is the fence. **MAX frees a buffer stream-ordered on its OWNING stream only.** A reader on any other stream races the allocator's reuse of the block, and that is not theoretical: the stress test reproduced corruption **40 times out of 40** before the fix, and **0 out of 40** after. Every allocation is now wrapped in a `_HolderOwner` proxy; `device_streams.record_use` records an event on the foreign stream, and the proxy's destructor enqueues a wait for that event on the owning stream *before* the release enqueues behind it. The free ends up ordered after every foreign reader, with no host synchronization anywhere.

### Everything goes through MAX — no vendor library

There is no `libcuda` binding in this PR. Streams, events, ordering and completion are all MAX APIs, so `device_streams.py` contains nothing CUDA-specific: a side stream is whatever extra stream MAX gives the device.

AMD is expected to work through exactly these abstractions, but it is **UNVERIFIED** — no AMD hardware was available. Only CUDA has been run. The module says so in its own docstring rather than claiming portability it has not earned.

### Two kinds of event, split by role

| | backing | used for |
|---|---|---|
| completion event | `max.driver.DeviceEvent` | `is_ready()` polling, `synchronize()`, `elapsed_time()` |
| fence event | Mojo `DeviceEvent` in `tensor_holder` | one stream waiting on an event another stream recorded earlier |

The split exists because no single API does both. The fence needs a wait, enqueued on the owning stream **arbitrarily later** (at destructor time), for an event recorded at a **point in time** on the foreign stream. `DeviceStream.wait_for(stream)` is not a substitute: it waits on the other stream's *current tail*, which over-fences and would require keeping the stream object alive. Python's `max.driver.DeviceEvent` has no wait-on-event at all. Mojo's `DeviceStream` has `enqueue_wait_for`, so two ~5-line functions in the `tensor_holder` extension carry it:

```mojo
fence_event_record(stream_ctx_ptr) -> FenceEvent   # create + record
fence_event_wait(stream_ctx_ptr, event)            # enqueue the wait
```

`FenceEvent` is a MAX `DeviceEvent` owned by a CPython object, so the AsyncRT refcount is handled by Mojo's own destructor when Python drops the last reference — **nothing crosses the ABI as a bare pointer and there is no manual retain/release to get wrong.** `tensor_holder` is compiled on first use of the device anyway, so this adds no build cost. It deliberately exposes no host wait: Mojo's event offers only `synchronize()`, and the Python event already covers that plus `is_ready()` and `elapsed_time()`.

### One addressing scheme

Streams are named end to end by their per-stream MAX `DeviceContext` pointer (`DeviceStream._device_context_ptr()`) — the same integer the kernel extensions already take for a context, because a MAX device context is a view over the device's stream set that differs only in which stream it submits to. The native `CUstream`/`hipStream_t` survives as `Stream.handle` for the one consumer that genuinely needs one: NCCL.

### Validation

Measured on an H100, against the reworked implementation:

- `fence_event_record` on a side stream + `fence_event_wait` on the default stream: **0** wrong elements with the fence, **33 554 432** wrong without it. The fence is real and load-bearing.
- A `copy_d2d` enqueued through a side stream's context is untouched by a default-stream sync (0.011 ms) and takes 0.70 ms to drain on the side stream — the context pointer really does address that stream.
- 1000 fence events created, waited on and dropped: no leak, no crash.
- `device._device_context_ptr() == device.default_stream._device_context_ptr()`, which a test now asserts so a future MAX change cannot quietly desynchronize the owning stream from the fence target.
- The free-race stress test (10 trials): reuse happens, corruption never does.

### Known trade-off: thread-safety

The completion worker polls events from its own thread and holder destructors run on whatever thread drops the last reference, so MAX objects are touched concurrently. MAX/AsyncRT does not document its thread-safety. The `libcuda` binding an earlier revision of this branch used could at least point at the driver API's documented guarantees; that argument is gone, and the exposure is accepted deliberately in exchange for a module that is not tied to one vendor. It is stated plainly in the module docstring rather than left implicit. MAX's `enqueue_host_func` / `wait_for_host_value` would let the polling be replaced by stream-signalled completion — out of scope here.

---

Slice of #421; independent. This is the backend's `record_stream` analog; the torch.Stream/Event PR builds `aten::record_stream` on it.

**Edit:** renamed `channels.py`/`Channel` to `device_streams.py`/`Stream` throughout to follow torch's own naming (`torch.cuda.Stream`) — `get_channel`→`get_stream`, `default_channel`→`default_stream`, `wait_channel`→`wait_stream` — and trimmed comments/docstrings down to non-obvious content. No behavior changed; `tests/test_mojo_streams.py` (1 GPU) and ruff are green.

https://claude.ai/code/session_017AbxaDeRuA7yywYEUuZV72

### DIFF
--- a/torch_mojo_backend/eager_kernels/tensor_holder.mojo
+++ b/torch_mojo_backend/eager_kernels/tensor_holder.mojo
@@ -22,7 +22,7 @@
 
 from std.os import abort
 from std.gpu import block_dim, block_idx, grid_dim, thread_idx
-from max.gpu.host import DeviceContext, DeviceBuffer, HostBuffer
+from max.gpu.host import DeviceContext, DeviceBuffer, DeviceEvent, HostBuffer
 from std.memory import memcpy
 from std.python import Python, PythonObject
 from std.python._cpython import PyObjectPtr, Py_ssize_t
@@ -404,6 +404,72 @@ def read_scalar(
 
 
 # ---------------------------------------------------------------------------
+# Fence events: cross-stream ordering for the stream-ordered free.
+#
+# A holder's release is stream-ordered on its OWNING stream only, so a reader
+# on another stream races the allocator's reuse of the block. The remedy
+# (mojo_device/device_streams.py, mojo_device/torch_mojo_tensor._HolderOwner) needs
+# exactly one primitive Python's `max.driver` does not expose: enqueue a wait
+# for an event recorded at a POINT IN TIME on one stream onto another stream
+# ARBITRARILY LATER — at destructor time, long after the recording stream has
+# moved on. `DeviceStream.wait_for(stream)` waits on the other stream's
+# current TAIL instead, which over-fences and would require keeping the
+# foreign stream object alive; `DeviceEvent` has no Python wait-on-event. Mojo
+# has both, so the two functions below carry it.
+#
+# `FenceEvent` is a MAX `DeviceEvent` owned by a CPython object: the AsyncRT
+# refcount is managed by Mojo's own destructor when Python drops the last
+# reference, so nothing crosses the ABI as a bare pointer and there is no
+# manual retain/release to get wrong. Streams are addressed by their
+# per-stream `DeviceContext` pointer (`DeviceStream._device_context_ptr()` on
+# the Python side), the same integer handle every other function here takes
+# for a device context.
+#
+# There is deliberately no host wait here: `DeviceEvent.synchronize()` is the
+# only other thing Mojo's event offers, and Python's `max.driver.DeviceEvent`
+# already provides that plus `is_ready()` and `elapsed_time()`, so host-side
+# completion is entirely a Python-side concern (mojo_device/device_streams.py).
+#
+# Dropping a FenceEvent immediately after enqueuing its waits is safe: the
+# driver defers the real destruction until the captured work completes.
+# `create_event`'s defaults (timing disabled, non-blocking sync) are what a
+# pure ordering event wants.
+# ---------------------------------------------------------------------------
+
+
+struct FenceEvent(Movable, Writable):
+    """Owns one MAX `DeviceEvent` for as long as Python holds this object."""
+
+    var event: DeviceEvent
+
+    def __init__(out self, event: DeviceEvent):
+        self.event = event
+
+    def write_to(self, mut writer: Some[Writer]):
+        # DeviceEvent is not Writable, so the reflection-derived default
+        # cannot be used for either of these.
+        writer.write("FenceEvent()")
+
+    def write_repr_to(self, mut writer: Some[Writer]):
+        writer.write("FenceEvent()")
+
+
+def fence_event_record(stream_ctx_ptr: PythonObject) raises -> PythonObject:
+    """Create an event and record it on `stream_ctx_ptr`'s stream."""
+    var ctx = _get_ctx(stream_ctx_ptr)
+    var event = ctx.create_event()
+    ctx.stream().record_event(event)
+    return PythonObject(alloc=FenceEvent(event^))
+
+
+def fence_event_wait(stream_ctx_ptr: PythonObject, event: PythonObject) raises:
+    """Order later work on `stream_ctx_ptr`'s stream after `event`."""
+    _get_ctx(stream_ctx_ptr).stream().enqueue_wait_for(
+        event.downcast_value_ptr[FenceEvent]()[].event
+    )
+
+
+# ---------------------------------------------------------------------------
 # CopyStrided: dst[coords] = src[coords] over an arbitrary-rank-<=8 index
 # space, with independent element strides on both sides (0-stride broadcast
 # reads included). This is the shared materialize primitive: .contiguous(),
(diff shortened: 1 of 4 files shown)


torch_mojo_backend/eager_kernels/tensor_holder.mojo
```diff
@@ -403,6 +403,72 @@ def read_scalar(
...
+#
+# There is deliberately no host wait here: `DeviceEvent.synchronize()` is the
+# only other thing Mojo's event offers, and Python's `max.driver.DeviceEvent`
+# already provides that plus `is_ready()` and `elapsed_time()`, so host-side
+# completion is entirely a Python-side concern (mojo_device/device_streams.py).
+#
+# Dropping a FenceEvent immediately after enqueuing its waits is safe: the
+# driver defers the real destruction until the captured work completes.
+# `create_event`'s defaults (timing disabled, non-blocking sync) are what a
+# pure ordering event wants.
```
gabrieldemarmiesse: This description is way too long,

torch_mojo_backend/mojo_device/device_streams.py
```diff
@@ -0,0 +1,202 @@
...
+
+1. Drain the kernel-call queue before ordering a side stream against the
+   default stream — a queued-but-unlaunched kernel is invisible to any
+   stream (`deferred_compile.drain()`).
+2. Hold a Python reference to every tensor a side stream reads or writes
+   until its work is known complete. Frees are stream-ordered on the
+   DEFAULT stream, so a buffer freed while a side stream still uses it is a
+   use-after-free.
+3. Anything the default stream must observe from a side stream needs an
+   explicit fence (event or host wait); nothing is implicit.
```
gabrieldemarmiesse: make this much shorter

gabrieldemarmiesse (changes requested):

torch_mojo_backend/eager_kernels/tensor_holder.mojo
gabrieldemarmiesse-ai-agent: Trimmed in d39e4f1 — the block is now ~10 lines / the module docstring ~18.

torch_mojo_backend/mojo_device/device_streams.py
gabrieldemarmiesse-ai-agent: Trimmed in d39e4f1 — the block is now ~10 lines / the module docstring ~18.

torch_mojo_backend/mojo_device/torch_mojo_tensor.py
```diff
@@ -60,6 +60,57 @@ def _holder_mod():
...
+        events = self._events
+        if not events:
+            return
+        try:
+            wait = self._wait
+            owner_ctx = self._owner_ctx
+            for event in events.values():
+                wait(owner_ctx, event)
+        except Exception:
+            pass  # interpreter teardown: the device context is going away too
```
gabrieldemarmiesse: this is too broad, we're going to swallow errors there

torch_mojo_backend/mojo_device/torch_mojo_tensor.py
```diff
@@ -60,6 +60,57 @@ def _holder_mod():
...
+        self._wait = None
+
+    def data_ptr(self) -> int:
+        return self._holder.data_ptr()
+
+    def get_nbytes(self) -> int:
+        return self._holder.get_nbytes()
+
+    def __getattr__(self, name: str) -> object:
+        return getattr(object.__getattribute__(self, "_holder"), name)
```
gabrieldemarmiesse: list exactly what attributes are accessible with getters rather than issuing a blanked gettatribute

torch_mojo_backend/mojo_device/torch_mojo_tensor.py
```diff
@@ -60,6 +60,57 @@ def _holder_mod():
...
+    ``tensor_holder.fence_event_wait`` at record time because ``__del__`` can
+    run during interpreter shutdown, when module globals are already cleared.
+    """
+
+    __slots__ = ("_holder", "_events", "_owner_ctx", "_wait")
+
+    def __init__(self, holder: object) -> None:
+        self._holder = holder
+        self._events = None  # {foreign stream ctx ptr: newest FenceEvent}
+        self._owner_ctx = 0
```
gabrieldemarmiesse: shouldn't this be None to avoid errors later on? Why 0?

gabrieldemarmiesse (changes requested):

torch_mojo_backend/mojo_device/torch_mojo_tensor.py
gabrieldemarmiesse-ai-agent: _owner_ctx now starts as None — with a garbage 0 the destructor would happily enqueue a wait on context pointer 0; None makes any misuse raise immediately. Fixed in the latest commit.

torch_mojo_backend/mojo_device/torch_mojo_tensor.py
gabrieldemarmiesse-ai-agent: Deleted the blanket __getattr__ entirely (bc06d2f) — a census showed the only attributes anything reaches through the owner are data_ptr() and get_nbytes(), and both were already explicit methods. Nothing else is delegated now.

torch_mojo_backend/mojo_device/torch_mojo_tensor.py
gabrieldemarmiesse-ai-agent: The except is gone (bc06d2f). __del__ now returns early only during interpreter finalization (sys.is_finalizing, captured at import since sys globals can be cleared by then); outside shutdown, failures surface via Python's unraisable hook instead of being swallowed.

PR #466, opened 2026-09-15
Move the native kernel cache to the user's cache directory

gabrieldemarmiesse:
The compiled shim, backend, op extensions and kernel variants were cached in-tree at eager_kernels/__mojocache__/native/, so removing the venv (or the checkout) threw away every build. They now default to platformdirs' user cache dir -- ~/.cache/torch-mojo-backend/native on Linux (XDG_CACHE_HOME honored), ~/Library/Caches/torch-mojo-backend/native on macOS -- and TORCH_MOJO_BACKEND_CACHE_DIR still overrides it. The cache is contents-addressed, so one directory serves every checkout on the box.

platformdirs was already installed transitively through mblack; declare it.

Docs that stated the path as current fact follow; the cluster notes now say to point TORCH_MOJO_BACKEND_CACHE_DIR at scratch rather than to put __mojocache__ there, since the default is now under $HOME.


pyproject.toml
```diff
@@ -12,6 +12,8 @@ requires-python = ">=3.11"
 dependencies = [
     "mojo==1.0",
     "max==26.5",
+    # The per-user cache directory the compiled native libraries live in.
+    "platformdirs",
```
gabrieldemarmiesse:
```suggestion
    "platformdirs",
```

PR #469, opened 2026-09-15
Pin host memory for real, so is_pinned() stops lying

gabrieldemarmiesse-ai-agent:
`Tensor.pin_memory()` silently returned ordinary pageable memory, and
`is_pinned()` was always `False`:

```python
>>> torch.arange(1024).pin_memory().is_pinned()
False
>>> torch.empty(16, pin_memory=True).is_pinned()
False
```

`MojoHooks::isPinnedPtr` returned `false` unconditionally and
`getPinnedMemoryAllocator()` handed back the plain CPU allocator.
`at::native::is_pinned` / `_pin_memory` are CompositeExplicitAutograd and route
straight through those two hooks, so no aten registration was involved — the
two hooks were the whole gap.

## What changed

The pinned allocator owns MAX `HostBuffer`s from `enqueue_create_host_buffer`
(page-locked), tracked in a base-sorted registry that `is_pinned` searches by
interval, so **interior** pointers answer correctly — the asynchronous-transfer
follow-up needs that to decide whether a direct DMA into a host address is safe.
Three entries appended to `TmbBackendHooks`; no op registration, no transfer
path touched.

Two upstream details drove the rest:

* **`GetCPUAllocatorMaybePinned`** (`EmptyTensor.cpp`) prefers CUDA's pinned
  allocator whenever `hasCUDA()`, while `at::accelerator::getAccelerator`
  prefers PrivateUse1 *unconditionally*. On a CUDA wheel the factory therefore
  allocates through CUDA and the query arrives here — we would answer "not
  pinned" about genuinely page-locked memory, and `pin_memory()`'s
  `if (self.is_pinned()) return self;` short-circuit would re-pin it through our
  allocator. `isPinnedPtr` falls through to CUDA so the answer is *true*, not
  merely *ours*.
* **`Context::isPinnedPtr`** short-circuits false while `init_[PrivateUse1]` is
  unset, and only `getPinnedMemoryAllocator` sets it — which the CUDA-wheel
  factory path never calls. So the first `is_pinned()` in a process answered
  false depending on call order. Registration now flips that guard
  (`lazyInitDevice`, present since v2.7.0, so it compiles on every supported
  torch).

A zero-byte request gets a null `DataPtr`, as the device allocator already does,
so both wheels agree with CUDA that a zero-byte tensor is not pinned — rather
than splitting on environment. `isPinnedPtr` refuses in a forked child before
touching either runtime, per the contract in `tmb_internal.h`.

## Tests

`tests/test_mojo_device.py` (both `mojo_device` params, so CI exercises them on
the MAX CPU device) covers the method and factory entry points agreeing,
zero-byte, strided round trips, interior pointers, a `DataLoader` with
`pin_memory=True`, a fresh-interpreter check that the *first* query is right,
and a `VmRSS` leak check over 256 MiB of pinned churn.
`tests/native/test_fork.py` covers the forked-child query.

Locally: `tests/test_mojo_device.py` 61 passed, plus the 4 CUDA-interop
failures that are **pre-existing on `main`** (verified in a clean worktree at
31ddc73) and unrelated to this change.

## Follow-up

A second PR builds asynchronous H2D/D2H on this registry: downloads that don't
drain the stream, and H2D from an already-pinned source that skips the staging
memcpy.


gabrieldemarmiesse-ai-agent:
Updated. Three things since the first push:

**All logic moved out of C++.** Per AGENTS.md, `native/csrc/` only holds what torch demands as a C++ class. The shim now forwards unconditionally: the zero-byte case, the registry lookup and the CUDA fall-through all live in `device.mojo`, and `tmb_cuda_is_pinned_ptr` is the single fact Mojo cannot answer for itself. The redundant `hasCUDA()` guard went with it, which also drops a `cudaGetDeviceCount` probe from every query.

**Corner-case tests**, from a contract-derived plan written without reading the implementation. Two bugs fell out:

- `torch.empty(..., device="mojo", pin_memory=True)` silently ignored the flag; it now raises `Only dense CPU tensors can be pinned`, as `at::detail::empty_cuda` does.
- **Both allocator deleters deadlocked in a forked child.** A child inherits `tmb_mutex` possibly locked by a thread that no longer exists, so freeing there hangs forever. This was already reachable on `main` for device memory — a forked child that drops any mojo tensor hangs — and newly reachable for pinned memory. Both now refuse to take the lock in a bad fork, consistent with what `tmb_internal.h` already promises for every other runtime entry.

The plan also corrected the spec I was working from: sparse tensors **can** be pinned upstream (by pinning their components), an empty *view* of a live pinned block is still pinned even though a zero-byte *allocation* is not, and provenance follows runtime CUDA availability rather than whether the wheel ships CUDA.

Locally: 562 passed. The only failures are the four pre-existing CUDA-interop ones, which fail identically on `main`.

PR #471, opened 2026-09-15
Asynchronous GPU downloads over pinned host memory

gabrieldemarmiesse-ai-agent:
GPU→CPU transfers synchronized unconditionally (`copy_to_host`'s docstring was
literally `"""Blocking D2H."""`), so `non_blocking=True` bought nothing and
CPU-offload pipelines lost their overlap. The headline case from the report now
works:

```python
downloaded = gpu_tensor.to("cpu", non_blocking=True)
# before: ~queue-draining, pageable      after: 0.04 ms, pinned, overlapped
```

`tests/test_mojo_device.py` carried an `xfail` naming exactly this; it is now a
passing test.

## What changed

**Downloads** skip the stream sync only into a live pinned block *we* allocated,
on the same device and stream. Everything else still blocks: pageable
destinations (CUDA's own rule — "For transfers from device memory to pageable
host memory, the function will return only once the copy has completed"),
CUDA-owned pinned memory, and blocks pinned against another mojo device. Those
last two are a deliberate safety margin over CUDA, which will happily DMA into
memory whose lifetime it cannot track; documented in `native_backend.md`.

**Uploads** from such a block DMA straight from the caller's memory instead of
paying a full-payload staging memcpy.

**`_to_copy` honours `non_blocking` and `pin_memory`.** It previously ignored
both, which is why `.to("cpu", non_blocking=True)` could never be async: upstream's
composite sets `pin_out = non_blocking && <accelerator> && cpu && strided` and
allocates a *pinned* destination, but our PrivateUse1 registration replaces that
composite and `_download_to_cpu` allocated pageable. We now mirror the upstream
condition. (`isAcceleratorExcluded` is present as far back as 2.7.0.)

**Deferred free.** A pinned block freed with a DMA in flight leaves the live
registry at once but is destroyed later, from a drain pass, once one completion
callback per recorded stream has fired — so a data loader dropping last
iteration's batch never unmaps memory the driver is still reading. Freeing never
synchronizes, which would destroy the very overlap this PR adds. Where MAX has no
host callbacks the fallback synchronizes that stream instead. The old `Staging`
box folded into this machinery rather than running a second one alongside.

## Bugs caught while building it

- **Stale-data race (silent corruption).** An async download followed by a
  *blocking* upload from the same buffer: the staging path memcpy'd on the host
  immediately, before the queued download had written it. Reproduced at
  1,048,576/1,048,576 elements wrong. A source in our registry now always takes
  the stream-ordered direct DMA, syncing when blocking — what CUDA does.
- **`.to()` ignored `non_blocking` entirely**, so the most idiomatic spelling
  never reached the new fast path.
- Host-conversion ordering, broadcasting, and lazy-negative handling, found by
  the new tests.

## Tests

~900 lines, from a contract-derived plan written *without* reading the
implementation — which is also what caught the `pin_out` gap above. Staged vs
direct uploads are distinguished by mutating the source after the call returns,
never by timing; ordering hazards assert values, not latency. Timing tests keep
the existing generous-margin structure for a shared GPU.

Original implementation validation reported 1210 passed, with four pre-existing
CUDA-interop failures also reproduced on `main`.

Updated against `main` (`49e8a64`), including #469 and #470. The merge preserves
peer-copy routing, exact integer casts, and staging-buffer ownership and
quarantine on failed uploads. The async download path now uses the shared
`is_cast_dtype` helper introduced on `main`.

Validation of the merge update:

- Transfer/pinned-memory selection in `tests/test_mojo_device.py`: 516 passed,
  630 skipped, 97 deselected.
- Dtype/layout/cross-device selection in `tests/native/test_data_movement.py`:
  21 passed, 225 skipped, 435 deselected.
- `uv run ty check` and `uvx pre-commit run --all-files`: passed.
- C++ shim and Mojo backend build successfully with `TORCH_MOJO_BACKEND_WERROR=1`.

GPU validation (2026-09-16), two Slurm H100 80 GB GPUs, PyTorch 2.11.0+cu128:

- Transfer/pinning/nonblocking tests: **1,133 passed, 12 skipped**. The one
  CUDA→Mojo interoperability failure reproduces identically on both the
  original PR (`fdce5c2`) and `main` (`49e8a64`).
- Cross-device/dtype/layout/clone tests: **267 passed**.
- Peer-copy fault-injection and lifetime tests: **15 passed**.
- **54 performance cases**, using actual CUPTI device time including DMA,
  alternating revision order and stock-PyTorch controls: no measured regression
  above the repository's 8% threshold after longer reruns of initial outliers.
  Worst median paired increase: **1.18%** over the original PR for its 48
  supported cases; **6.69%** over main for six peer-copy cases (the original
  PR did not support peer copies). Tested float32/bfloat16 and 4,096,
  281,673, and 16,777,216 elements.
- Four initial timing outliers were rerun for 24 rounds, with 64 profiled
  calls/burst and 512 host-timed calls/burst on a fixed CPU core; none
  reproduced. These reruns were within +0.25% GPU time and +1.5% host-call
  time of their controls.
- Repository clone/cast benchmarks: **6 passed, 2 skipped**. This environment
  has no recorded entries, so the direct comparisons above establish the
  regression result. No baselines were updated.
- Driver permissions prevented clock locking. Results are limited to the
  measured H100 cases. No further production-code changes were needed.


gabrieldemarmiesse-ai-agent:
Updated after a review pass. Four fixes:

**A performance regression I had introduced.** Making a blocking upload from a pinned source synchronize was right, but the `synchronize()` landed on the *staging* path too — where it is wrong. That path already takes a synchronous host memcpy snapshot, so the caller can reuse its buffer the moment the call returns; blocking semantics were already satisfied, which is why the pre-PR code never synchronized there. Every internal caller that omits `non_blocking` started draining the stream. Median of six calls behind a 300 ms stream hold:

| Operation | Before | After |
|---|---:|---:|
| Large-int64 `arange` | 300.27 ms | 0.10 ms |
| Pageable `copy_` | 300.29 ms | 0.06 ms |
| Pageable `.to()` | 300.28 ms | 0.06 ms |

**`pin_out` now matches upstream exactly.** I had used `pin_memory or non_blocking`. Upstream's `options.pinned_memory(pin_out)` *overwrites* the caller's `pin_memory` on this path, so `_to_copy(device="cpu", pin_memory=True, non_blocking=False)` returns pageable memory there — and a test of mine had enshrined the divergence. Checked against v2.7.0 and 2.11; both agree.

**A failed pinned allocation no longer breaks `.to("cpu")`.** Page-locked memory is a scarce OS resource, so exhausting it is an expected condition, not a bug. It now falls back to the pageable blocking route instead of raising where the call used to succeed.

**Flaky timing assertions fixed**, and both transfer timing tests now alternate legs and take a median, per AGENTS.md's benchmark notes — the upload one never had that either.

1316 passed locally; the only failures are the four pre-existing CUDA-interop ones, identical on `main`.

PR #472, opened 2026-09-15
Memory accounting and device introspection for torch.mojo

gabrieldemarmiesse-ai-agent:
`torch.mojo` had no memory APIs and no device introspection, and Inductor's
`MojoInterface.memory_allocated()` was a hardcoded `0`. Memory-aware
placement, batch sizing and monitoring had nothing to read.

```python
torch.mojo.memory_allocated()         # 4194304
torch.mojo.max_memory_allocated()     # 4194304
torch.mojo.memory_stats()["allocation.all.current"]
torch.mojo.mem_get_info()             # (free, total) from MAX
torch.mojo.get_device_properties()    # name, cc, SM count, warp size, ...
torch.accelerator.memory_allocated()  # the same number, same allocator
```

## What made this work

One fact decided the design: `torch.accelerator`'s entire memory surface
goes through `at::getDeviceAllocator(PrivateUse1)`, which `dynamic_cast`s
whatever `c10::GetAllocator` returns to `c10::DeviceAllocator` and
`TORCH_INTERNAL_ASSERT`s if the cast fails. `MojoAllocator` derived from
plain `c10::Allocator`, so every `torch.accelerator.memory_*` call threw.
Changing that base class is what lights up the whole surface, and the
interface is byte-identical in torch 2.10 and 2.11, so no version guards.

Following that through, `torch.mojo`'s functions read torch's own
`torch._C._accelerator_*` bindings rather than a parallel set of accessors.
`torch.mojo.memory_allocated() == torch.accelerator.memory_allocated()` is
then true by construction rather than by coincidence.

Everything MAX can answer is answered from Mojo -- `get_memory_info`,
`name`/`api`/`arch_name`, `get_attribute` -- with no libcuda or libamdhip
probing, so a CPU-only torch wheel still gets real numbers. The C++ side
gained no decision logic: it forwards the six `DeviceAllocator` overrides to
Mojo hooks and marshals a flat `int64` array into torch's `DeviceStats`.
Counters, retry/OOM accounting and the per-device property cache all live in
`device.mojo`.

## Honest about what the numbers mean

`reserved_bytes` equals `allocated_bytes` because there is no caching
allocator of ours -- MAX owns the arena. `memory_reserved() -
memory_allocated()` is structurally zero here, not a measurement of slack.
The `small_pool`/`large_pool`, `segment`, `active*` and `inactive_split*`
breakdowns are structurally zero for the same reason: this backend has none
of those concepts. `empty_cache()` releases completed pinned host staging
buffers and cannot return device memory to the OS, because MAX exposes no
trim API.

`mem_get_info()` reports MAX's arena budget, not the card: on the H100 here
it says 73216 MiB where `nvidia-smi` says 81559 MiB. `free` tracks
allocations exactly -- a 2 GiB tensor moves it by 2048 MiB -- so it is the
right number for batch sizing, but it is not `cudaMemGetInfo`. The docs say
so rather than papering over it.

`num_alloc_retries` and `num_ooms` are real signals nobody had: MAX's
allocator can fail transiently on a full arena, and `_create_retry` already
drained and retried without anyone being able to see it happen.

## Two pre-existing defects the tests caught

* `storage.resize_(-1)` reached the allocator as `SIZE_MAX` and silently
  produced a storage reporting `nbytes() == -1`. Reproduced on `main`; now
  rejected at the ABI boundary before MAX or the counters are touched.
* `new_strided` allocated through `tmb_empty_strided` and only then built
  the `T` handle, which rejects unsupported ScalarTypes -- stranding the
  handle and its storage. No op currently reaches it (the dtype is refused
  a layer earlier), so this is latent, but it is a leak by construction.

## Tests

`tests/native/test_memory.py`, 226 passing across 47 test functions. The
plan behind them was written by a second agent that was **forbidden from
reading this repo** and derived 49 cases from upstream PyTorch's contract
alone -- which is how the two defects above surfaced, and how the cold-state
early returns in `torch/accelerator/memory.py` and the
`allocation.*`-vs-`allocated.*` key-spelling discrepancy in upstream's own
docstrings got covered.

Beyond the obvious counters: views/slices/`expand` charging once, autograd
saved buffers, Python reference cycles, `set_`/storage resize, non-default
streams, frees from another thread, per-device isolation across device
switches, meta and fake tensors charging nothing, cumulative counters past
2^32, overflow requests failing before they touch accounting, fork safety,
and every `torch.accelerator.*` call agreeing with its `torch.mojo.*`
counterpart.

Second-GPU isolation skips on this box (one GPU). Inductor's
`DeviceProperties.create` is untestable here for the pre-existing reason
that `enable_inductor()` refuses on a CUDA-enabled torch wheel.

Local: 226 passed / 6 skipped for the new suite, 500 passed / 1 skipped
across bringup, fork, factories, data movement and streams; `ty` clean.


gabrieldemarmiesse-ai-agent:
Review pass (astra, adversarial): **no production-code defects** — counter concurrency under the shim mutex, the flat-array C ABI (slot counts, short buffers, oversized names), the 2.9/2.10 version gates against tags 2.7–2.13, AGENTS.md compliance, and fork safety (including forking while another thread held the mutex) all came back clean.

It did find two tests that passed against a *broken* implementation, and proved it by mutation rather than by argument:

* `test_empty_cache_preserves_live_storage` passed whether `empty_cache` was a no-op or synchronized the device — it read device counters, which by design exclude the host staging buffers `empty_cache` actually releases.
* `test_accelerator_agrees` allocated 8193 bytes and never asserted they appeared; every comparison was between two wrappers over the same call, so zeroing both surfaces passed.

Both are rewritten with independent oracles — current `VmRSS` for staging release (MAX's host cache disabled in isolated workers, since it otherwise hides the release), alternating-median timing for the non-synchronizing claim, and a test-owned byte ledger for the accelerator agreement. They now fail against all three mutations and pass against the real implementation:

```
noop         REJECTED: completed staging was not released: 0 of 67108864 bytes
synchronize  REJECTED: empty_cache waited for pending work: median 0.314s, synchronize 0.314s
zero_stats   REJECTED: assert api.memory_allocated() == base_bytes + live_bytes
real         PASSED
```

One claim is deliberately narrowed rather than faked: that *pending* staging buffers survive is not observable through the public API, and the docstring says so instead of asserting it decoratively.

228 passed / 6 skipped; timing test 10/10 stable across five runs.

PR #475, opened 2026-09-16
Pick the ptxas that fits the driver and the GPU, and say so when none does

gabrieldemarmiesse-ai-agent:
Every kernel ends in ptxas, and which ptxas runs is pinned between two bounds
that pull in opposite directions: the driver sets an upper one (a CUDA 13
cubin needs r580; on r570 it loads as `CUDA_ERROR_INVALID_IMAGE`) and the GPU
sets a lower one (sm_110 needs CUDA 13, and an assembler that does not know
the architecture fails the build with `Value 'sm_90a' is not defined for
option 'gpu-name'`). Until now we hardcoded one answer — the CUDA 12.8 wheel —
which is right for most machines and silently wrong for the rest, with no
message that said so.

This picks the assembler from what the machine actually is, and says why when
it cannot.

## What it does

`MODULAR_NVPTX_COMPILER_PATH` is now chosen in two steps:

* **at import**, from the driver alone. `cuDriverGetVersion` is callable
  before `cuInit`, so it creates no context, touches no GPU and is safe
  before `max` loads and before any fork.
* **at `register_mojo_devices()`**, from the architecture, which only an
  initialized driver can answer — and still before the first build.

Candidates are the `nvidia-cuda-nvcc*` wheels (`nvidia/cuda_nvcc/bin/ptxas`
for cu12, `nvidia/cu13/bin/ptxas` for the CUDA 13 wheel), `torch/bin/ptxas`,
Triton's, `$CUDA_HOME`, `$CUDA_PATH`, `$PATH` and `/usr/local/cuda*`. Each is
asked its release (`--version`) and which architectures it targets
(`--help`) — never assumed, because the lower bound is not monotonic: CUDA 13
added sm_110 and *dropped* sm_101 and everything below sm_75.

Among the ones that fit, the wheel `pyproject.toml` pins wins. It is what the
kernels are tested and tuned with, and a torch upgrade shipping a newer ptxas
must not silently move a working machine onto it. A switch away from it is
traced on stderr, so a bug report mentions it.

A `MODULAR_NVPTX_COMPILER_PATH` the user set is never overridden — only
checked, and explained if it cannot work here. The pick is also marked in the
environment (`TORCH_MOJO_BACKEND_PTXAS_AUTO`), because a child process
inherits the environment and nothing else: without it every torchrun rank and
every Inductor compile worker would read our own value as a setting of the
user's and refuse to move off it.

Nothing runs at import on a machine with no NVIDIA driver: the wheel's path is
set without asking any binary its version, so an AMD or CPU-only import cannot
be held up by a hung toolkit wrapper on a network mount.

## What the user sees

Before, on an r570 box whose assembler was a CUDA 13 one:

```
RuntimeError: At max/mojo/max/gpu/host/_device_context_extras.mojo:168:17:
CUDA call failed: CUDA_ERROR_INVALID_IMAGE (device kernel image is invalid)
```

— at the first op, after every build had succeeded. Now, at registration:

```
MODULAR_NVPTX_COMPILER_PATH=.../torch/bin/ptxas cannot be used here: cubins
need driver r580 or newer; this driver supports CUDA 12.8.
Unset it and .../nvidia/cuda_nvcc/bin/ptxas (CUDA 12.8) is used instead,
which works on this machine.

  driver    570.211.01 (supports CUDA 12.8)
  gpu 0     NVIDIA H100 80GB HBM3 (sm_90)

  ptxas found
    [  no] .../torch/bin/ptxas
           CUDA 13.0, from MODULAR_NVPTX_COMPILER_PATH -- cubins need driver r580 or newer; this driver supports CUDA 12.8
    [pick] .../nvidia/cuda_nvcc/bin/ptxas
           CUDA 12.8, from nvidia-cuda-nvcc-cu12 wheel
    [  ok] .../triton/backends/nvidia/bin/ptxas
           CUDA 12.8, from triton wheel
    [  ok] /usr/local/cuda-12.8/bin/ptxas
           CUDA 12.8, from system CUDA

  `torch-mojo-backend ptxas` prints this table at any time.
```

When nothing on the machine fits, the same table ends with the wheel to
install — `pip install "nvidia-cuda-nvcc-cu12==12.8.*"`; a `--no-deps` line
plus what it costs when the GPU needs a release newer than this package's own
pin; "update the NVIDIA driver" when no wheel can bridge the gap; and, for two
GPUs with no assembler in common, that they cannot share a process. `torch-mojo-backend ptxas` prints it on demand (it works with
no GPU and no CUDA at all), a `mojo build` failure mentioning ptxas points at
it, and `TORCH_MOJO_BACKEND_PTXAS_CHECK=0` turns the refusal into a warning
for a machine whose rules we got wrong.

## Validated on an H100 node (driver 570.211.01, CUDA 12.8)

| setting | before | after |
|---|---|---|
| default | works | works, unchanged |
| a CUDA 13.0 ptxas | `CUDA_ERROR_INVALID_IMAGE` at the first op | refused at registration, with the table |
| a path that is not there | `mojo build` dump at the first op | refused at registration, "no such file" |
| a CUDA 11.8 ptxas | `ptxas fatal : Value 'sm_90a' is not defined`, inside a build dump | refused at registration, "CUDA 11.8 cannot target sm_90a" |

54 unit tests, written as the machines they describe (an r570 box with a
CUDA 13 torch wheel, a Blackwell needing a newer assembler than the pin, a
Volta on a driver whose newest assembler dropped it). The advice tables are
held up against the real binaries installed here: that test is what caught
CUDA 13 dropping sm_101.

AMD and Apple are untouched — the whole path is skipped without an NVIDIA
driver.

## Review

A second review (GPT-6 Astra, via the Codex CLI) found six issues, all fixed
here and all covered by tests; the retry latch, the inheritance case and both
advice defects have tests confirmed red against the pre-fix commit:

1. `check()` latched itself before it could raise, so a retried registration —
   which this project supports on purpose — skipped the check and failed later
   on a message that explains nothing.
2. Provenance was module state, so a spawned rank read our own inherited pick
   as the user's and refused to re-select for its GPU.
3. The advice for a GPU newer than the pin said `uv add
   "nvidia-cuda-nvcc-cu12>=12.9"`, which cannot resolve against this package's
   own `==12.8.*`. It now says `pip install --no-deps`, and what that costs.
4. Import ran every discovered assembler's `--version` even with no NVIDIA
   driver present.
5. `ARCH_MIN_CUDA[90]` held sm_90's minimum (11.8) when what we require is
   sm_90a (12.0); sm_52, sm_53 and sm_72 were off by a release too.
6. A test demanded that every assembler discovered on the machine runs, which
   a broken system CUDA install would fail.

https://claude.ai/code/session_01EH1sPzkLWmihhP7Z6NMfvn


gabrieldemarmiesse-ai-agent:
Re-validated on an H100 node (par2dc5-ai-prd-cl02s03dgx30, driver 570.211.01 / CUDA 12.8) at the final commit `6dba831`, i.e. after the review fixes:

| `MODULAR_NVPTX_COMPILER_PATH` | result |
|---|---|
| unset (the wheel's 12.8 is picked) | `tensor([2., 2., 2., 2.])` — unchanged |
| torch's CUDA 13.0 ptxas | `PtxasError` at registration: *cubins need driver r580 or newer; this driver supports CUDA 12.8* |
| a CUDA 11.8 ptxas | `PtxasError` at registration: *CUDA 11.8 cannot target sm_90a* |
| `/nonexistent/ptxas` | `PtxasError` at registration: *no such file* |
| `/nonexistent/ptxas` + `TORCH_MOJO_BACKEND_PTXAS_CHECK=0` | warning only, then the old `mojo build` failure — now ending in a pointer to `torch-mojo-backend ptxas` |

Each refusal prints the full candidate table and, when nothing on the machine fits, the wheel to install.

Tests on that node: `tests/test_ptxas_selection.py`, `tests/test_ptxas_default.py`, `tests/test_cli.py` — 58 passed; `tests/native/test_bringup.py`, `tests/native/test_register_retry.py` — 29 passed. Earlier run on the same tree: `test_loader.py` and `test_fork.py` included, 52 passed.

AMD and Apple are unmeasured; the path is a no-op there (no `libcuda.so.1`, so no probing and no check).

PR #478, opened 2026-09-17
Train FSDP2 GPT-2 XL on the mojo device at stock-CUDA speed (NCCL and MojoCCL)

gabrieldemarmiesse-ai-agent:
FSDP2 can train sharded models on the Mojo device with vendor NCCL or MojoCCL, and on 2 and 8 H100s it now runs at stock-CUDA speed. Both stacks turned out to be host-bound (GPU compute-stream time is ~110 ms of a ~224 ms step), so the work is host cost per op and per collective, not kernels:

- Process group: an event-backed `Work` per collective instead of a device Future (all_gather_into_tensor 105 → 71 µs in-model).
- Boxed-kernel adapter: argument conversions decided once per schema; a tensor's metadata read in one call; the shim is compiled at -O2 (`009379e`, droppable). Registering ATen's own view kernels moved to its own PR, #482.
- Kernel launch cache keyed by the kernel function's compile-time identity: a warm launch formats and allocates nothing; `KernelCall` uses inline storage; the loader no longer zeroes a 4 KiB error buffer per launch.
- Op layer: one fast route for the common elementwise case; `mul_`/`sub_` write in place; `addmm` adds its bias inside the op; a pre-existing `add_`-on-offset-view CUDA misaligned-address fault is fixed in the elementwise kernel.
- MojoCCL single node: reduce-scatter is one push+reduce kernel per call (was ~122 chunked all-reduces and a host synchronize); all-gather reads its contribution once; AVG scales each contribution (NCCL PreMulSum).
- Earlier in this PR: batched FSDP split copies and FP32→BF16 parameter copies, adjacent foreach copies, direct in-place optimizer ops, fused tanh backward, Hopper FP32 row copies and sqrt.

Throughput (GPT-2 XL, BF16 blocks / FP32 root and reductions, batch 1/GPU, sequence 1024, eager AdamW and clipping with `foreach=False`; medians over 6–10 synchronized 10-step windows in palindromic stack order, nothing else on the nodes; details in `docs/gpt2_fsdp2_throughput.md`):

| GPUs | Stock CUDA + NCCL | Mojo + NCCL | Mojo + MojoCCL |
|---|---:|---:|---:|
| 2 | 9,062.8 | 9,141.7 (100.9%) | 9,186.1 (101.4%) |
| 8 | 35,582.6 | 34,648.6 (97.4%) | 34,399.4 (96.7%) |
| 16 (2 nodes), 18 windows | 70,819.0 | 68,692.2 (97.0%) | 67,173.7 (94.9%) |

The 16-GPU row is an 18-window palindromic run (three palindromes) on one node pair; two-node windows spread +-7% on every stack, and shorter runs have read NCCL at 95-100% and MojoCCL at 93-97%, so 2-3% differences there are at the noise floor. MojoCCL's multi-node reduce-scatter is hierarchical (node-local push+reduce, one RDMA shard exchange, inbox sum; was a chunked all-reduce placeholder at 21%), its all-gather sends one contribution per NIC and releases the RDMA exchange from inside the gather, and a one-block rank gate absorbs start-barrier skew; grids are fitted end to end because a collective CTA resident on an SM blocks the persistent GEMMs (NCCL runs 16 CTAs). MojoCCL is within 2% of NCCL as the collective library at 16 ranks. Isolated at FSDP2's sizes: all-gather 0.7-0.9x NCCL; reduce-scatter 1.10x (transformer block, 48 of the 49 calls per step) and 1.28x (root) with the new streaming kernel `rs_stream.mojo` (per-piece flag handoff instead of grid barriers; was 1.39x/1.48x). The root's remainder is measured, not modelled: deleting its two HBM passes leaves 1.17x, so a node-local ring that fuses load-add-store plus a faster push are both needed. Small messages (<= 1 MiB) stay 1.5-2x NCCL (fixed barrier cost; FSDP2 never issues them). The remaining 16-rank gap to stock is the Mojo device (present with NCCL too), under investigation. Remaining 8-GPU gap is the optimizer phase's host time (MAX's stream-ordered frees of two temporaries per parameter, ~1.5 µs each). Every mojo 8-GPU launch shows one ~10% slower window, unexplained.

Validation: two-rank distributed suite (NCCL and MojoCCL: collectives, DDP parity, stream ordering, stress, abort, both FSDP2 modes, chunked reduce-scatter, AVG overflow) 15/15; targeted native suites (bringup, binary, unary, matmul, loader, data movement, reductions, foreach, nn, attention, random parity, factories, composed, stream ordering, generator, compare) 3,663 passed, 0 failed; ruff, ty, env-var/monkeypatching/annotation invariants and benchmark coverage pass; GPT-2 124M prints the same five-step losses under both collective libraries as before. Every chunk was reviewed by a second model (GPT-6 Astra) and the findings fixed. AMD is cross-compiled but unmeasured.

Rebased state: `main` (after #481, one backend library for all op bodies) is merged in; on that tree the two-rank distributed suite (15/15) and the targeted native suites (2,655 passed, 0 failed) pass on H100.


gabrieldemarmiesse-ai-agent:
Merged `main` (now carrying #484, the one-root Mojo layout) into this branch, so it applies on top of the new tree:

- The seven Mojo files this PR adds moved with their packages: `rs_fused.mojo`, `rs_multinode.mojo`, `rs_stream.mojo` under `tmb/ccl/`, `tanh_backward.mojo` under `tmb/kernels/activation_backward/`, `batched_copy_cast.mojo`, `batched_copy_rows.mojo`, `copy_segments.mojo` under `tmb/kernels/data_movement/`. Their imports and every import this PR added elsewhere are absolute `tmb.` paths; the family strings it added (`data_movement_ops`, `optimizer_ops`, `activation_backward_ops`, `logic_ops`) use the new names.
- `kernel_call_probe.mojo` and the multinode `geometry_test.mojo` build with the single `-I torch_mojo_backend/mojo`; `test_mojoccl_reduce_scatter_layout.py` reads `tmb/ccl/rs_stream.mojo` and `tmb/ccl/entry.mojo`.
- One latent bug fixed on the way: `roi/entry.mojo` still passed a string key to `_enqueue_cached` at two call sites after this PR changed its signature, so the `roi` and `deform_conv` families did not compile (nothing in CI builds them; they need a GPU).
- `shim_dispatch.cpp` keeps both #482's `tmb_count_op_call` and this PR's arena comment. `test_boxed_adapter_error_kinds` asserted that a bad `view` shape error names `aten::view`; since #482 ATen's own kernel serves `view`, so the test now expects ATen's message.

Checked on the merged tree: every family entry, libmojoccl and the backend build under `--Werror`; `tests/test_mojo_imports.py` and the other GPU-free suites pass; GPU runs of the families this PR touches (loader, data movement, foreach, composed, bringup, checkpoint, torchvision ops, attention, and a sample of binary/matmul/unary) — 1685 passed, 233 skipped, 3 xfailed, 2 xpassed, 0 failed after that test fix (56 min on one H100).

gabrieldemarmiesse-ai-agent:
Split out of this PR so each part can be reviewed on its own (all against `main`, all on the one-root Mojo layout):

- #487 — boxed adapter: conversions decided once per schema, tensor metadata in one shim call, shim at -O2 (5 commits, independent)
- #485 — event-backed `Work` and per-call device validation in the process group (3 commits, independent)
- #486 — DCP checkpoint staging for Mojo tensors when the loader picked another accelerator (the checkpoint half of one commit, independent)
- #490 — the eager op fast paths and the kernels behind them, plus the `cat.out` dtype widening from this PR's first commit that the cat-cast kernel relies on (10 commits, independent)
- #491 — the kernel launch cache and heap-free `KernelCall` (8 commits, stacked on #490)

What stays here: MojoCCL (single-node reduce-scatter/all-gather, AVG pre-scaling, the hierarchical and streaming multi-node reduce-scatter), FSDP2 support, the demo, the throughput notes, and the collective-ordering half of the checkpoint commit. Once the five land, merging `main` into this branch shrinks it to exactly that.

gabrieldemarmiesse-ai-agent:
#490 (the op fast paths) is itself split three ways now: #495 (binary op host path), #496 (Hopper FP32 row copies and sqrt) and #490 (batched copies for split / cat / `_foreach_copy_`, the BF16 cat cast, the fused tanh backward). #491 is stacked on the three.

